PR #65, opened 2022-02-14
Use argparse.ArgumentParser.parse_known_args

blindmatrix:
bitbake crash if you supply arguments gn-utils.inc doesn't support (but https://github.com/flutter/engine/blob/main/tools/gn does). This PR simply ignore unknown arguments.


blindmatrix:
Too soon, wasn't done yet

PR #66, opened 2022-02-14
Use argparse.ArgumentParser.parse_known_args

blindmatrix:
bitbake crash if you supply arguments gn-utils.inc doesn't support (but https://github.com/flutter/engine/blob/main/tools/gn does). This PR simply ignore unknown arguments.

### DIFF
--- a/conf/include/gn-utils.inc
+++ b/conf/include/gn-utils.inc
@@ -211,7 +211,7 @@ def parse_args(d):
     parser.add_argument('--tsan', default=False, action='store_true')
     parser.add_argument('--ubsan', default=False, action='store_true')
 
-    return parser.parse_args(args)
+    return parser.parse_known_args(args)[0]
 
 def get_out_dir(d):
     """Gets output directory based on the Package Configuration args."""


jwinarske:
@blindmatrix I'm not clear on the scenario that triggers crash.  Can you provide example input on what this PR protects against?

blindmatrix:
> @blindmatrix I'm not clear on the scenario that triggers crash. Can you provide example input on what this PR protects against?

Sure thing!

I'm building this for a device shipped without X11 which mean GLFW wont compile, so I have a file, `flutter-engine-release_%.bbappend`, like this
```
GN_ARGS_append = " --no-build-glfw-shell --no-build-embedder-examples"
```

to prevent it from building an example embedder which require GLFW.

`gn` shipped with the `flutter-engine` support `--no-build-embedder-examples` while the shorter version you have in `gn-utils.inc` doesn't. If parser.parse_args(args) see unknown arguments a short "helpful" text will be printed and the process will terminate with error-code 2. This in turn will case bitbake to crash in a such a way that leave some bitbake child processes running too, resulting in me having to `pkill -f bitbake` before being able to run any other bitbake command again.

jwinarske:
@blindmatrix Cool.  Yeah I have some strong feelings about the defaulting to building glfw, let alone having dep on x11 :)

I'll update the parser to master gn, and add `disable-embedder-examples` to package config.  So at least it will be available.

I know engine stable branch does not have the hot fixes to disable embedder-examples by default.  I have open flutter issue on this.

PR #77, opened 2022-03-01
Feature/gclient fetcher

WAKazuyoshiAkiyama:
Summary:

- add gn.bbclass/gn.py, which is the bitbake fetcher for gcleint. By using this bbclass, flutter-engine can store its source codes in DL_DIR. This is more suitable with Yocto style and will make the result of builds more reusable.

- Change SRC_URI of flutter-sdk from the git repo to the tarball.
  - When major updating, flutter-sdk will overwrite its history. Due to this, the build may be unstable.
  - By using the release tarball, this issue can be avoided.

- add .gitignore
  - this is created to ignore *.pyc of gn.py


jwinarske:
@WAKazuyoshiAkiyama 
Can you rebase against `main`?  CI should be ready to run in a few hours.

WAKazuyoshiAkiyama:
@jwinarske I rebased this branch.

jwinarske:
This engine build is past thirty minutes.  Any idea why?  How long should it take with an empty cache?
https://github.com/meta-flutter/meta-flutter/runs/5402080716?check_suite_focus=true

jwinarske:
A clean build of the engine currently takes ~6 minutes.  One example:
https://github.com/meta-flutter/meta-flutter/runs/5400935106?check_suite_focus=true

jwinarske:
@WAKazuyoshiAkiyama The machine backing up the CI jobs on meta-flutter is a 48 core + 128GB RAM machine with 1GB internet bandwidth.  Your proposed solution is very slow.  It took 1 hour and 8 minutes to build one instance of the engine.  Currently in meta-flutter using gclient it complete engine step in ~6 minutes, which includes a cleansstate, fetch, and build each time.

My initial thought on flutter/gclient LTS was to parse out the DEPS file in base repo, then generate a recipe similar to how Rust bitbake cargo works.  It would be slightly slower than gclient implementation, but it should be much faster than your proposed scheme.  So a pre-processing script that creates an inc file.  The outer recipes using this could be named something like: 
flutter-engine-lts-debug_2.10.2.bb
flutter-engine-lts-release_2.10.2.bb
flutter-engine-lts-profile_2.10.2.bb

If there was a minimal difference in build time I would drop depot_tools/gclient usage altogether.

jwinarske:
@WAKazuyoshiAkiyama This tool in theory can convert DEPS to a parsable list:
https://chromium.googlesource.com/chromium/tools/deps2git

jwinarske:
@WAKazuyoshiAkiyama Actually something better.  `gclient` leaves behind a json file of all the paths/repos called `.gclient_entries`.  It's written in the same folder as `.gclient` and `.cipd`.

jwinarske:
Yocto defaults branch as `master`.  A number of the repos do not have a `master` branch, rather only `main`; most likely renamed.  I'm finding there is trial and error to find which repos default to `main`, then updating python output for some repos with `branch=main`.  It might be good to have the script validate each repo and determine which requires `branch=main`.

jwinarske:
Working branch for experiment.  Target CI job is linux-dummy.
https://github.com/meta-flutter/meta-flutter/blob/jw/lts/recipes-graphics/flutter-engine/flutter-engine-debug_2.11.0-0.1.pre.bb

WAKazuyoshiAkiyama:
> Currently in meta-flutter using gclient it complete engine step in ~6 minutes, which includes a cleansstate, fetch, and build each time.

At least, in my environment, even if gn.bbclass is not used, engine step spends 1+ hour, so I'm not sure why the case without gn.bbclass is so faster.

The way to use gclinet.py should not be so much different between gclinet,py in do_patch and gn.bbclass....

WAKazuyoshiAkiyama:
> similar to how Rust bitbake cargo works

I like this idea.

From my understanding, since cipd's SRCREV is specific, so CIPD fetcher class needs to be created. Is it correct?

jwinarske:
> 
> From my understanding, since cipd's SRCREV is specific, so CIPD fetcher class needs to be created. Is it correct?

Correct

WAKazuyoshiAkiyama:
@jwinarske 
I tried to create cipd fetcher class.

But, the error of git fetching happened.

```
ERROR: flutter-engine-debug-git-r0 do_fetch: Fetcher failure: Unable to find revision fa166c29f0407be21acebb394d194a38c00593cb in branch main even from upstream
ERROR: flutter-engine-debug-git-r0 do_fetch: Fetcher failure for URL: 'git://dart.googlesource.com/sdk.git;protocol=https;destsuffix=src/third_party/dart;branch=main;name=dart_googlesource_com_sdk_git'. Unable to fetch URL from any source.
```

The git revisions of flutter-engine dependencies might be unstable, so this idea might be difficult to use.

I created flutter-engine-src.inc based on https://github.com/meta-flutter/meta-flutter/blob/jw/lts/recipes-graphics/flutter-engine/flutter-engine-debug_2.11.0-0.1.pre.bb.

But, the entries of cipd:// has the extra `_` and `/` in SRC_URI and SRC_REV, so I removed them.

```
cipd:/// -> cipd://
name=_XXX -> name=XXX

SRCREV__XXX -> SRCREV_XXX
```

jwinarske:
@WAKazuyoshiAkiyama 
I have output pulling files now.  The issue is that Yocto wants explicit branch of commit, and it defaults to master.  I had to implement logic in the [script](https://gist.github.com/jwinarske/cd45ca42be2e6091295dfa5587ee9daa) to query repo.  In the case of the freetype2 repo, the commit doesn't exist on the server, so I drop it.  In the case of libwebp the tags changed to have a 'v' prefix.  So I added some logic to get first right justified match.

Removing the `cipd://` entries + freetype2 entry, the CI job completes file sync in ~5 minutes.  So this is good.

**Process to generate inc file**
Navigate to engine folder with .gclient
`gclient revinfo --output-json=gclient_entries.json`
move gclient_entries.json to same folder as gclient_bitbake.py
`python gclient_bitbake.py > ${FLUTTER_SDK_TAG}.inc`

In the version I tried `2.11.0-0.1.pre` it has an invalid reference to freetype2.  Commit cannot be found in any branch.  My suspicion is that this path isn't used in the build, and it's been this way for a while.
```
    "src/third_party/freetype2": {
        "url": "https://flutter.googlesource.com/third_party/freetype2",
        "rev": "1f03c1b2d7f2ae832a4fbe9d12bd96c3c15bbece"
    },
```

For the CIPD implementation I was thinking it should model after file fetcher, and use same properties.

Example
```
cipd://flutter/web/canvaskit_bundle;destsuffix=src/third_party/web_dependenciesflutter_web_canvaskit_bundle
```
* the first section is the CIPD path
* destsuffix is location under ${S} where CIPD package gets extracted to
* it _**should**_ have an md5 property as well similar to downloading a file

It's just like a https file download that uses CIPD transport on top of curl.  Generic CIPD download support.

jwinarske:
@WAKazuyoshiAkiyama 
I have the script working well, and it runs on machine with many cores.  Execution time to validate all the SRC_URI is around 7 minutes using 20 cores.

Looking at cipd in depot tools, it's a bash script.  It attempts to pull an updated client before each runs.  It pulls cipd executable from google storage.   It should be using cipd client that was built via yocto build, without an update shim.

The source for luci infra is located here:  https://github.com/luci/luci-go

To build cipd tool
```
export PATH=/home/dev/depot_tools:$PATH
export STAGING=`pwd`/.tmp
fetch infra
cd infra/go
eval `./env.py`
cd src/go.chromium.org/luci
mkdir $STAGING/bin
go build -v -o "$STAGING/bin/cipd" go.chromium.org/luci/cipd/client/cmd/cipd
$STAGING/bin/cipd help -advanced
```

I added a recipe that builds cipd here: https://github.com/meta-flutter/meta-flutter/blob/jw/lts/recipes-devtools/luci-go/luci-go-native_git.bb

The runtime path for flutter-engine is `${STAGING_DIR_NATIVE}/usr/bin/cipd`

jwinarske:
@WAKazuyoshiAkiyama 

My branch has a working cipd fetcher that uses the `luci-go-native` cipd binary.  Fetching for `gclient-entries.inc` appears to be 100% functional.  Performance fetching CIPD packages is very good.  Less than a minute on my system to pull all the packages.

I'm now hitting what looks like an incorrect repo commit, or overlay ordering issue.  I'm done for the weekend, and will look at it tomorrow (US).
```
| ERROR at //third_party/dart/runtime/bin/BUILD.gn:5:1: Can't load input file.
[250](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:250)
| import("../../build/config/gclient_args.gni")
[251](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:251)
| ^-------------------------------------------
[252](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:252)
| Unable to load:
[253](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:253)
|   /__w/meta-flutter/yocto-dunfell/poky/build/tmp/work/core2-64-poky-linux/flutter-engine-debug/2.11.0-0.1.pre-r0/git/third_party/dart/build/config/gclient_args.gni
[254](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:254)
| I also checked in the secondary tree for:
[255](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:255)
|   /__w/meta-flutter/yocto-dunfell/poky/build/tmp/work/core2-64-poky-linux/flutter-engine-debug/2.11.0-0.1.pre-r0/git/build/secondary/third_party/dart/build/config/gclient_args.gni
[256](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:256)
| See //flutter/shell/testing/BUILD.gn:35:5: which caused the file to be included.
[257](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:257)
|     "//third_party/dart/runtime/bin:dart_io_api",
[258](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:258)
|     ^-------------------------------------------
[259](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:259)
| Generating GN files in: out/linux_debug_x64
[250](https://github.com/meta-flutter/meta-flutter/runs/5440886778?check_suite_focus=true#step:9:250)
```

Something we need to look at is adding md5 support to the CIPD fetcher.

I added md5 entries for each cipd package to gclient-entries.inc.

WAKazuyoshiAkiyama:
@jwinarske 
I don't understand how to use this [script](https://gist.github.com/jwinarske/cd45ca42be2e6091295dfa5587ee9daa).

This script is not executed in the Yocto and executed before Yocto build to generate *.inc file?

If so, how should `cipd;//` entries be handled?

I'm not sure what your current point is and what you want to do.

WAKazuyoshiAkiyama:
@jwinarske 
Could you share the branch of luci-go? I would like to check your work about luci-go.

jwinarske:
@WAKazuyoshiAkiyama 
This [script](https://github.com/meta-flutter/meta-flutter/blob/jw/lts/scripts/gclient_bitbake.py)  is run from folder with .gclient file, which creates snapshot for Yocto.  The output - gclient-entries.inc is used by a flutter-engine recipe.

What I have in my branch, it pulls all the cipd:// entries and extracts then.

Regarding luci-go you can Google for the source repo and documents.  It's very light on documentation.  There's a lot that can be done with that (infra/luci-go) source tree.  Cipd client is just one of many luci tools.  I used my script to create the luci-go-native recipe.  The version I grabbed snapshot of was main.  You run gclient sync after checking out repo that holds the root DEPS.  This will enforce all dependent deps.

Next is to pull in asset download using the save from google storage python script (depot tools), wrapping it with a Google storage fetcher gs://.

This is needed for SDK download, as well as some actions in root DEPS "hooks".

You can start on that.

The other item is md5 on cipd fetcher is needed.

WAKazuyoshiAkiyama:
For cipd fetcher class I implemented, this cannot handle multiple cipd:// entries, so I need to fix this...

WAKazuyoshiAkiyama:
I tried to use [gclient_bitbake.py](https://gist.github.com/jwinarske/cd45ca42be2e6091295dfa5587ee9daa) in flutter 2.10.3, but, somewhy, SRCREV of engine cannot be generated.

```
SRC_URI += "\
    git://github.com/flutter/buildroot.git;protocol=https;destsuffix=git/;name=github_com_flutter_buildroot_git;name=g
ithub_com_flutter_buildroot_git \
    cipd://fuchsia/third_party/clang/linux-amd64;destsuffix=git/buildtools/linux-x64/clang/;name=fuchsia_third_party_c
lang_linux_amd64 \
    git://github.com/flutter/engine.git;protocol=https;destsuffix=git/flutter/;name=github_com_flutter_engine_git \
    cipd://gn/gn/linux-amd64;destsuffix=git/flutter/third_party/gn/;name=gn_gn_linux_amd64 \
[...]
SRCREV_github_com_flutter_buildroot_git = "4660f4a166979c64ff56f6b856b44ccc1dba5c63"
SRCREV_fuchsia_third_party_clang_linux_amd64 = "Fn7lDYhKDAwbGQ2SOL_Anwt8fzO1Yho7UjpoS9Hv8N8C"
SRCREV_github_com_flutter_engine_git = ""
SRCREV_gn_gn_linux_amd64 = "git_revision:b79031308cc878488202beb99883ec1f2efd9a6d"
SRCREV_fuchsia_sdk_core_linux_amd64 = "DS-3oCidC7xGrdGByHplsRyY8Err4sguVaV0CfHsn6sC"
[...]
```

How can I get a correct gclient-entris.inc?

jwinarske:
@WAKazuyoshiAkiyama That indicates the json file has value null for the revision.  Either update json with value, or add revision after the fact.  I could check this state and cd into src/flutter to return the commit hash.  Will look into that.

Also there is something odd happening with libwebp in the script.  I had it working before.  The gist is that the `1.2.0` revision should be `v1.2.0`.  Just modify it after the fact until I get it working again.

jwinarske:
@WAKazuyoshiAkiyama The cipd fetcher appears to handle multiple entries.  Check my branch for the changes.

WAKazuyoshiAkiyama:
@jwinarske I have already fixed my cipd fetcher in my local. (self.basecmd -> ud.basecmd)
But, you have already implemented some fetchers, so can I rebase this branch according to your branch?

WAKazuyoshiAkiyama:
@jwinarske 
> That indicates the json file has value null for the revision. Either update json with value, or add revision after the fact. I could check this state and cd into src/flutter to return the commit hash. Will look into that.

I have already  added the srcrev of engine manually, but the following error happened.

```
ERROR: flutter-engine-debug-2.10.3-r0 do_fetch: Fetcher failure: Unable to find revision bd539267b42051b0da3d16ffa8f48949dce8aa8f in branch main even from upstream
ERROR: flutter-engine-debug-2.10.3-r0 do_fetch: Fetcher failure for URL: 'git://github.com/flutter/engine.git;protocol=https;destsuffix=git/flutter/;branch=main;name=github_com_flutter_engine_git'. Unable to fetch URL from any source.
```

But, the revision `bd539267b42051b0da3d16ffa8f48949dce8aa8f` appears in the engine repo.

```
$ cd src/flutter
$ git show
commit bd539267b42051b0da3d16ffa8f48949dce8aa8f (grafted, HEAD, origin/flutter-2.8-candidate.16)
Author: Kevin Chisholm <kevinjchisholm@google.com>
Date:   Mon Feb 28 16:09:52 2022 -0600

    [flutter_releases] Flutter stable 2.10.3 Engine Cherrypicks (#31721)

    * Conditionally call FlutterViewDestroyOverlaySurfaces (#31464)

    * Fix PhysicalShapeLayer paint bounds with clipping disabled (#31656)

    * Fix PhysicalShapeLayer paint bounds with clipping disabled

    * Add missing SkRect initialization

    * 'add branch flutter-2.8-candidate.16 to enabled_branches in .ci.yaml'

    * remove branch ref

    Co-authored-by: Emmanuel Garcia <egarciad@google.com>
    Co-authored-by: Matej Knopp <matej.knopp@gmail.com>
```

WAKazuyoshiAkiyama:
I could check the fetch failure of freetype2 you said.
And, in this PR, I will try to accelerate gn.bbclass fetching.

You are working on the idea like cargo.bbclass in [that PR](https://github.com/meta-flutter/meta-flutter/pull/80/files#), so I will try another option.

jwinarske:
> 
> I don't understand why the error happens...
@WAKazuyoshiAkiyama This happens as the rev version for the root project in the json file is null.

I updated the script.  Changes:
* correctly identifies when requested commit is not found
* using path from DEPS fills in source revision in case where  json root project "rev" == null.
https://github.com/meta-flutter/meta-flutter/blob/jw/lts/scripts/gclient_bitbake.py

In case for freetype2 it would expect it would fail to fetch, but it doesn't break the build.  I haven't examined the log yet.

WAKazuyoshiAkiyama:
@jwinarske 
> And, in this PR, I will try to accelerate gn.bbclass fetching.

> You are working on the idea like cargo.bbclass in [that PR](https://github.com/meta-flutter/meta-flutter/pull/80/files#), so I will try another option.

Is it OK?

The idea like cargo.bbckass might have the issue revisions of dependecies may be unstable(this issue might be resolved by the improvement of the script), so if gn.bbclass becomes much faster, gn.bbclass is better, I think.

I think it is better that I try to make gn.bbclass faster while you try the idea like cargo.bbckass.

WAKazuyoshiAkiyama:
The procedure to create tarball with "tar -cJf" may take long, so I will try to change the procedure to create tarlball with multi thread.

jwinarske:
@WAKazuyoshiAkiyama
If we need LTS today then your gclient snapshot approach would work.  I think the client pre-process script approach will be used by the Yocto community for general gclient support.  It's simple and effective, and is easy to update to a new release.  I'll notify the mailing list tomorrow.

The time to beat for your gclient snapshot approach is ten minutes when running on my machine.

Cloud time is very expensive for high core count vCPU.  So I would use the most effecient solution.

WAKazuyoshiAkiyama:
@jwinarske 
Yes, I agree your approach will have a better performance.
But, I'm worry that flutter-engine upstream fluency changes the revisions of dependencies in a same version e.g. due to overwrite histroeis.
If such overwriting happens, the maintenance of flutter-engine LTS will become more difficult.
If such overwriting fluency happens, the approach of gclient snapshot will be more suitable even if that performance is lower.

WAKazuyoshiAkiyama:
@jwinarske 
I updated gn.bbclass (lib/gn.py) to make do_fetch faster.

After building flutter-engine 2.10.2, it takes about 6m to build flutter-engine 2.10.1.

```
$ time bitbake flutter-engine-debug -c fetch
real    5m56.239s
user    0m0.500s
sys     0m0.104s
```

Withtout a previous result in DL_DIR, do_fetch may take 40+m in my environment(12 core).

WAKazuyoshiAkiyama:
After updaing gn.bbclass (lib/gn.py), the directory to execute `gclient sync` are shared if the version of engine is changed.
But, the tarball generated by do_fetch is different between the versions of engine.

```
$ ls /workdir/download/gn
https__github.com_flutter_engine.git
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2.done
https__github.com_flutter_engine.git-ab46186b246f5a36bd1f3f295d14a43abb1e2f38.tar.bz2
https__github.com_flutter_engine.git-ab46186b246f5a36bd1f3f295d14a43abb1e2f38.tar.bz2.done
```

WAKazuyoshiAkiyama:
Overloading the `unpack` method and using pbzip2, do_unpack will be made faster, but, other bitbake fetcher doesn't do that, so I hesitate to overload the `unpack` method.

WAKazuyoshiAkiyama:
I added the specific unpack method for the gclient fetcher.
Due to this, do_unpack becomes made much faster.

```
$ time bitbake flutter-engine-debug -c unpack
real    1m54.932s
user    0m0.983s
sys     0m0.156s
```

Before adding this unpack method, do_unpack took 15+m.

WAKazuyoshiAkiyama:
@jwinarske Could you check the new implementation of gn.py in this repo's CI?
After second or later build, the time to build will become shorter enough.

WAKazuyoshiAkiyama:
@jwinarske 
Could you run CI?

WAKazuyoshiAkiyama:
@jwinarske 
Come to think of it, I think that when gclient sync was run, gclient did not simply fetch the dependent sources, but also executed a hook and fetched the Dart SDK, etc.
Is this kind of processing supported by [your PR](https://github.com/meta-flutter/meta-flutter/pull/80)?
If it is not supported, then you may need to run gclient runhook after getting a series of sources.

jwinarske:
@WAKazuyoshiAkiyama Yes.  I have a new task getting added to the inc file.  I'm working through the details on that right now.  In order to handle the conditionals in hooks and deps it requires come on the fly code compiling.  I am running the script on a number of other gclient projects (v8, Skia, etc).  Each uses the conditionals a little different.  When I am at a breaking point I will run your jobs.

jwinarske:
@WAKazuyoshiAkiyama This error is due to having desktop linux enabled.  I explicitly turn this off in dunfell branch.
https://github.com/meta-flutter/meta-flutter/actions/runs/1954631184

WAKazuyoshiAkiyama:
@jwinarske 
> This error is due to having desktop linux enabled. I explicitly turn this off in dunfell branch.
> https://github.com/meta-flutter/meta-flutter/actions/runs/1954631184

From my understanding, linux-desktop is disabled in do_compile of flutter-sdk in even also this branch.
So, this is the issue of CI?

> Mismatch of VM between SDK and Engine:
> https://github.com/meta-flutter/meta-flutter/runs/5474458756?check_suite_focus=true

I'm not sure why this error happens in this target only.
The revision of upstrean might be updated.
I'll check this error, and I may set a default value to FLUTTER_SDK_TAG if required.

WAKazuyoshiAkiyama:
> From my understanding, linux-desktop is disabled in do_compile of flutter-sdk in even also this branch.
So, this is the issue of CI?

in do_compile_class-native and do_compile_class-nativesdk, linux-desktop is not disabled, so I will make do_compie common in class-taget. class-native and class-nativesdk.

jwinarske:
@WAKazuyoshiAkiyama 
To rule out the build I would need to wipe the sstate and build.  Each build runs a cleansstate for each flutter build, to avoid anything strange.  Unclear what the sstate does when versions change.  There test would be to run the full set on clean everything.  I do this before a release, make sure everything is still working; cold boot smoke test...

The tag mismatch, not sure.  It shouldn't happen as it's explicitly defined for both.  The same variable is used for both SDK and engine.

I do have concerns about disk space getting eaten up per build.  I rarely clear the dl dir, and it's common to all the build.  I would need to manage that somehow.  What are the options there?

WAKazuyoshiAkiyama:
@jwinarske 
> I do have concerns about disk space getting eaten up per build. I rarely clear the dl dir, and it's common to all the build. I would need to manage that somehow. What are the options there?

Can I remove all of tarball generated when cleanall is executed?

Or, just executing `rm -rf ${DL_DIR}/gn/*flutter_engine*` will work.

And also, I can fix the version of flutter to e.g. 2.10.3.

jwinarske:
@WAKazuyoshiAkiyama 
distclean should also delete the dl cache file.
As long as you filter what you're deleting it should be good.

WAKazuyoshiAkiyama:
@jwinarske 
Currently, the gclient fetcher remove the tarball of current revision, i.e. the gclient fetcher doesn't remove the tarball already generated.

- Ex)
  - If there are the following tarbal in DL_DIR
    - DL_DIR/gn/https__github.com_flutter_engine.git-bd539267b42051b0da3d16ffa8f48949dce8aa8f.tar.bz2
    - DL_DIR/gn/https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2
  - the current engine revision is bd539267b42051b0da3d16ffa8f48949dce8aa8f.
  - In this case, even if cleanall/cleandist is executed, `DL_DIR/gn/https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2` will not be executed.

I feel it is not good to remove the different version's tarball, but I can change the impl to remove all of generated tarballs.

WAKazuyoshiAkiyama:
The error of nativesdk-flutter-sdk didn't happen in my local environment.

For the error that flutter-sdk hash is different between flutter-app and flutter-engine, in order to avoid that error, I added the inc file to set a default value of FLUTTER_SDK_TAG.

This is suitable for LTS support.

WAKazuyoshiAkiyama:
I found the root cause of https://github.com/meta-flutter/meta-flutter/runs/5474458756?check_suite_focus=true.
The logic to get the revision if AUTOREV is set is different between flutter-sdk and flutter-engine, so sometimes, this error will happen.
So, I made the logic common between them.

jwinarske:
@WAKazuyoshiAkiyama 
Regarding not removing prior versions, I agree.  We don't want that to happen.

Looks like remaining item is disabling Linux desktop to make SDK build happy.

Build times look good 👍

jwinarske:
@WAKazuyoshiAkiyama Update on my LTS solution:

1. freetype2 issue is still outstanding
I found something odd with gclient and freetype2.  I submitted issue hoping someone can answer my question.
https://github.com/flutter/flutter/issues/99841
gclient is doing this somehow.  It's not clear there is a git hook implementation doing this, or what.

2. Replacing freetyp2 with revision of head (for now) CI jobs are passing.  do_run_hooks gets dynamically generated in the latest [script](https://github.com/meta-flutter/meta-flutter/blob/jw/lts/scripts/gclient_bitbake.py).  I observed one engine debug build (cleansstate + build) which was 2 min 43 seconds.  That is by far the record for a clean engine build on my machine.

3. Once behavior with freetype2 is understand and accommodated, then I will look into running a batch job to convert all stable tags in flutter repo and add to meta-flutter.  Then use of FLUTTER_SDK_TAG would directly select the engine inc file.  Engine inc files will be located in conf/include.

jwinarske:
@WAKazuyoshiAkiyama 
I have changed hook script `python3 src/flutter/tools/download_dart_sdk.py --fail-loudly` to be a https SRC_URI entry.
Unless the build itself pulls artifacts during the build, I believe my PR is LTS compatible.

WAKazuyoshiAkiyama:
@jwinarske 
I fixed the error of invalid deps for linux-desktop.
flutter-sdk stored in storage.googleapis.com has some artifacts in `bin/cache/artifacts`, and they cause the error.
Maybe, those artifacts may be generated through something like release tests of Google, but, since they are unnecessary, I removed them in do_install.

Could you run CI?

jwinarske:
@WAKazuyoshiAkiyama 

My PR is passing all CI now.

The minimum change for your last change is:
https://github.com/meta-flutter/meta-flutter/pull/80/files#diff-78e6ed211e5f49ec21d6edd9971d3b7cfc89625c801896beae1d388dbcec979dR74
Deleting the whole cache would not prevent build, but will most likely re-trigger download during build.

We still need to solve problem of the Dart pub package manager.  Packages referenced in project yml are dynamically pulled at build, if they are not present in the pub package manager.  I'm thinking the solution here is to parse the application yml and manually add cache entries via a custom fetcher.  If we don't then these get added dynamically at build time (not LTS).

When I build flutter SDK I specify PUB_CACHE to be under ${S}.  I added this as the default pub cache gets created under user directory; host contamination.  In the case of the engine build it updates pub cache `offline` using:
```
src/third_party/dart/tools/sdks/dart-sdk/bin/pub get --offline
```
It's not clear if all the items the engine build consumes, are actually present in the cache to start.  We should be able to isolate that easily with desktop build environment.  I'm think they are not.

Note in the current scheme for meta-flutter when application is built PUB_CACHE is not set, so it's defaulting to user directory; host contamination and not LTS.  The host contamination can be resolved with export PUB_CACHE in do_compile of flutter app class.

When an app is built, it will install all the packages referenced in `pubspec.yml`, and install them into the pub package manager; location is determined via environmental variable PUB_CACHE.  To work around this we are going to need to add a fetcher to add flutter packages to the pub package manager.  In addition we will need to be careful how/when PUB_CACHE is set.

Once pub package manager is resolved we should run Wireshark on the subnet of a build machine and monitor the traffic.  If no pulls are seen, then we are LTS.

jwinarske:
@WAKazuyoshiAkiyama 
Dart can be run from any directory to get offline packages using the following pattern:
```
PUB_CACHE="my/path" dart pub get --directory=./flutter/examples/hello_world    
PUB_CACHE="my/path" dart pub get --directory=./flutter/examples/hello_world   --offline
```
If --offline is used in first pass I see dependency errors on some projects.

I believe the default behavior is `--no-offline`.

```
Usage: dart pub get <subcommand> [arguments...]
-h, --help               Print this usage information.
    --[no-]offline       Use cached packages instead of accessing the network.
-n, --dry-run            Report what dependencies would change but don't change any.
    --[no-]precompile    Build executables in immediate dependencies.
-C, --directory=<dir>    Run this in the directory<dir>.
```
So for LTS we need to include this in some format for a given app.

jwinarske:
@WAKazuyoshiAkiyama I just merged support for PUB_CACHE cache archive/restore for Flutter App support.  This doesn't include any pub caching for the engine recipe.

WAKazuyoshiAkiyama:
@jwinarske 
I understand the issue of PUB_CACHE for LTS.

From my understanding, flutter-engine has a similar issue.
When executing the hooks of gclient, some dart deps(pub cache) will be fetched in flutter-engine build.
And, the version of dart deps are not fixed in pubspec.yaml of flutter-engine.

At least, I failed to build the engine because the dart deps were updated at remote and the miss dependencies happened.
Then, I resolved the issue by adding pubspec.lock before executing `gclient runhooks`.

- `gclient sync --nohooks --noprehooks`
- add pubspec.lock to fetched source codes
- `gclient runhooks`

Shall I add some scripts to generate pubspec.lock for the engine and support pubspec.lock for the engine recipe?
If the version of flutter supported by meta-flutter is fixed on each revisions, the way to use pubspec.lock will work.

WAKazuyoshiAkiyama:
> Shall I add some scripts to generate pubspec.lock for the engine and support pubspec.lock for the engine recipe?
If the version of flutter supported by meta-flutter is fixed on each revisions, the way to use pubspec.lock will work.

I still don't think this is a good idea. In the way of gclient fetcher, it is better to get just dart deps and take its snapshot.

I will try to change PUB_CACHE dir in the engine recipe.

WAKazuyoshiAkiyama:
I tried to set PUB_CACHE env in lib/gn.py and classes/gn.bbclass, but, it seems to make no sense.
The hooks of flutter-engine might not be related to `pub get`...

jwinarske:
@WAKazuyoshiAkiyama 
I ran a config with `--ide=json`.  It provides a JSON project suitable for parsing.  I haven't had a chance to review it completely yet, but it does reference package lists explicitly for Dart projects.

WAKazuyoshiAkiyama:
By the way, do your PR use vpython? vpython generates some directories on $HOME e.g. ~/.vpython-root, so to avoid the contamination of the user's envrionment, we need to consider how to handle vpython,

WAKazuyoshiAkiyama:
@jwinarske 
https://github.com/meta-flutter/meta-flutter/pull/80 is ready to be merged?
To improve do_fetch and make it more stable to build flutter-engine, this PR or https://github.com/meta-flutter/meta-flutter/pull/80 should be merged, and if https://github.com/meta-flutter/meta-flutter/pull/80 can be merged, I will close this PR.
However, if https://github.com/meta-flutter/meta-flutter/pull/80 is difficult to be merged now, I need to ready to merge this PR ASAP, I think.
So, I would like to know your plan.

jwinarske:
@WAKazuyoshiAkiyama 
80 is ready although I'm still on the fence.  I am leaning towards your PR at the moment.  It would involve less maintenance, which I like, and build times are optimal.  Every stable release for 80 would require intervention.

For this PR there are a couple of minor changes needed:

The sync should happen in ${WORKDIR}, then extraction could be skipped, and it wouldn't need to be deleted in DL_DIR.  

I'm interested in the font recipe addition.  I grep'd the engine and SDK, without a reference to the font referenced as default.

WAKazuyoshiAkiyama:
@jwinarske 
> The sync should happen in ${WORKDIR}, then extraction could be skipped, and it wouldn't need to be deleted in DL_DIR.

I'm sorry, could you tell me how to improve concretely?
Do you mean do_unpack should be skipped?

> I'm interested in the font recipe addition. I grep'd the engine and SDK, without a reference to the font referenced as default.

Previously, when I ran the app not specified any fonts, if liberation-fonts package is not installed, the app cannot run.

Can I change the dependencies to liberation-fonts to RRECOMMENDS?

WAKazuyoshiAkiyama:
Even if PUB_CACHE and VPYTHON_VIRTUALENV_ROOT are specified in lib/gn.py like the following, any directories will not be generated in DL_DIR.

```
diff --git a/classes/gn.bbclass b/classes/gn.bbclass
index fd09fcf..ea67253 100644
--- a/classes/gn.bbclass
+++ b/classes/gn.bbclass
@@ -37,6 +37,7 @@ do_configure_prepend() {
     export HTTP_PROXY=${HTTP_PROXY}
     export HTTPS_PROXY=${HTTPS_PROXY}
     export PATH=${DEPOT_TOOLS}:${DEPOT_TOOLS}/${PYTHON2_PATH}:${PATH}
+    export PUB_CACHE=${S}/.pub-cache
     export DEPOT_TOOLS_UPDATE=0
     export GCLIENT_PY3=0
 }
@@ -47,6 +48,7 @@ do_compile_prepend() {
     export HTTP_PROXY=${HTTP_PROXY}
     export HTTPS_PROXY=${HTTPS_PROXY}
     export PATH=${DEPOT_TOOLS}:${DEPOT_TOOLS}/${PYTHON2_PATH}:${PATH}
+    export PUB_CACHE=${S}/.pub-cache
     export DEPOT_TOOLS_UPDATE=0
     export GCLIENT_PY3=0
 }
diff --git a/lib/gn.py b/lib/gn.py
index aac87ff..597bd02 100644
--- a/lib/gn.py
+++ b/lib/gn.py
@@ -64,10 +64,14 @@ class GN(FetchMethod):

         ud.basecmd = "export PATH=\"%s:%s:${PATH}\"; export DEPOT_TOOLS_UPDATE=0; export GCLIENT_PY3=0; \
             export CURL_CA_BUNDLE=%s; \
+            export PUB_CACHE=%s; \
+            export VPYTHON_VIRTUALENV_ROOT=%s; \
             gclient.py config --spec '%s' && \
             gclient.py sync %s --revision %s %s -v" % (
                 depot_tools_path, os.path.join(depot_tools_path, python2_path),
                 d.getVar('CURL_CA_BUNDLE'),
+                os.path.join(ud.syncpath, '.pub-cache'),
+                os.path.join(ud.syncpath, '.vpython-root'),
                 gclient_config,
                 sync_opt, srcrev, d.getVar('PARALLEL_MAKE'))
```

Therefore, setting PUB_CACHE and VPYTHON_VIRTUALENV_ROOT make no sense.

And, when GIT_CACHE_PATH was specified, the error happened in do_fetch, so this environment is not available.

So, I have no addition for the current revision.

WAKazuyoshiAkiyama:
> The sync should happen in ${WORKDIR}, then extraction could be skipped, and it wouldn't need to be deleted in DL_DIR.

In the current implementation, the sync directory for flutter-engine is reused when flutter-engine's version is changed.
This reusing help acceleration do_fetch much.

The following directory will be generated for any version.

```
/workdir/download/gn/https__github.com_flutter_engine.git
```

However, an artifact actually used in do_unpack is a generated tarball like the following.

```
$ ls /workdir/download/gn/*.tar.bz2
/workdir/download/gn/https__github.com_flutter_engine.git-bd539267b42051b0da3d16ffa8f48949dce8aa8f.tar.bz2
```

Dividing the working directory for gclient sync and the artifacts for do_unpack, building an old version of flutter-engine will become easier.

For example, after flutter-engine 2.10.2 is built, there are the following artifacts under DL_DIR/gn.

```
$ ls /workdir/download/gn/
https__github.com_flutter_engine.git
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2.done
```

And then, after flutter-engine 2.10.3 is build, there are the following artifacts under DL_DIR/gn.

```
https__github.com_flutter_engine.git
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2.done
https__github.com_flutter_engine.git-bd539267b42051b0da3d16ffa8f48949dce8aa8f.tar.bz2
https__github.com_flutter_engine.git-bd539267b42051b0da3d16ffa8f48949dce8aa8f.tar.bz2.done
```

And then, when building flutter-engine 2.10.2 again, `https__github.com_flutter_engine.git` is the state of version 2.10.3, but, since there is the tarball of flutter-engine 2.10.2 (`https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2`), `gclient sync` will be skipped and `https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2` will be used.

If `gclient sync` is directly executed in WORK_DIR, when changing the version of flutter-engine, `gclient sync` needs to be executed every time.
So, the build time will be increased.

I think the current way is better than executing directly `gclient sync` in WORK_DIR.

jwinarske:
@WAKazuyoshiAkiyama i understand.  My point is that it should use ${WORKDIR} opposed to a potentially orphan folder that consumes space.  The argument that it does a Delta sync faster versus an absolute sync, yes it would.  However the preference would be to have source tree in a  guaranteed state.  

The other concern is around depot_tools.  It will verify latest version prior to running, meaning a network connection is required, and the far end service needs to be functional.  My preference is to factor out the depot_tools dependency.  Why cipd wrapper script checks version everytime is a big question.  Does the backend interface change that often, or are they using it to gather use metrics similar to autoninja.  My preference is to factor out the depot tools dependency as PR 80 has done.  I believe this approach follows the spirit of the Yocto project more closely, as it quantifies the unknowns in an explicit way.

WAKazuyoshiAkiyama:
@jwinarske 
> My point is that it should use ${WORKDIR} opposed to a potentially orphan folder that consumes space. 

There is a usecase to share DL_DIR between develpers, e,g, with using SOURCE_MIRROR_URL.
So, I believe the souce code of flutter-engine should be placed in DL_DIR.

> The argument that it does a Delta sync faster versus an absolute sync, yes it would. However the preference would be to have source tree in a guaranteed state.

Actually, in the initial implementation of this PR, the sync dir is different between versions of flutter-engine like the following.

```
$ ls /workdir/download/gn
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2
https__github.com_flutter_engine.git-a83ed0e5e3b9cd2b5e2f07ef31c72f43c55e93b7.tar.bz2.done
https__github.com_flutter_engine.git-bd539267b42051b0da3d16ffa8f48949dce8aa8f
https__github.com_flutter_engine.git-bd539267b42051b0da3d16ffa8f48949dce8aa8f.tar.bz2
https__github.com_flutter_engine.git-bd539267b42051b0da3d16ffa8f48949dce8aa8f.tar.bz2.done
```

So, I can change the sync dir per versions of flutter-engine.
Shall I do so?

> I believe this approach follows the spirit of the Yocto project more closely,

I agree, but, the gclient fetcher firmly depends on gclient in depot_tools, so I cannot remove the dependency to depot_tools.

So, I think, to avoid some unclear issues like the following, patching depot_tools and disabling unnecessary network accesses are the only way.

>  It will verify latest version prior to running, meaning a network connection is required, and the far end service needs to be functional.

PR #80, opened 2022-03-04
LTS support

jwinarske:
-LTS build model
 removes depot_tools recipe
 each change in FLUTTER_SDK_TAG requires running the
 scripts/gclient_bitbake.py to generate conf/include/flutter-engine_%.inc.
 LTS support requires minor change to conf/include/flutter-engine_%.inc
 for using Dart SDK prebuilt via http fetcher.
    
-Removal of depot_tools
 No longer required
    

jwinarske:
Needs
1. gclient hook implementation.  To be called from do_configure:prepend.
2. gs:// fetcher for pulling google storage assets

jwinarske:
@WAKazuyoshiAkiyama
I had first successful engine build on this PR.  4 minutes and 11 seconds for debug build.  That is an all time record.  That includes building the Dart SDK, opposed to pre-built.

Missing parts were a gclient_gn_args_file needed to be created, and hooks that needed to be run.

The hooks are a concern, as if one of them pulls an artifact we need to capture these with a fetcher so they get cached to DL.

PR #89, opened 2022-03-21
gn: Rename gn.bbclass to avoid collision

rzr:
GN tool is not only specific to flutter,
Until it's upstreamed into a shared layer
let's rename it to avoid collision
with other projects using gn (ie: oniro)

BTW, may I suggest to later rename it to gclient too,
to prevent ambiguity with gn buid system.

Relate-to: https://gitlab.eclipse.org/eclipse/oniro-core/oniro/-/merge_requests/4
Forwarded: https://github.com/meta-flutter/meta-flutter/pulls?q=author%3Arzr
Signed-off-by: Philippe Coval <philippe.coval@huawei.com>

### DIFF
--- a/classes/gn_flutter.bbclass
+++ b/classes/gn_flutter.bbclass
@@ -1,0 +1,49 @@
+# Copyright (c) 2021-2022 Woven Alpha, Inc
+
+python () {
+    import gn
+    bb.fetch2.methods.append(gn.GN())
+}
+
+DEPENDS += " \
+    ca-certificates-native \
+    curl-native \
+    depot-tools-native \
+    pbzip2-native \
+    tar-native \
+"
+
+inherit python3native
+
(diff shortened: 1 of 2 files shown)


jwinarske:
@rzr This is a good point.  I will have changes made in Dunfell then pushed to Honister.

rzr:
feel free to pick change , I can also open MR to dunfell

jwinarske:
@WAKazuyoshiAkiyama We need to rename gn.  Can you create PR for this?

WAKazuyoshiAkiyama:
OK

WAKazuyoshiAkiyama:
I will create PR for dunfell tomorrow.

jwinarske:
@rzr cherry-picked from dunfell

Issue #115, opened 2022-05-19
[Linux Plugin Test App] runs in Debug, not AOT


jwinarske:
The Linux Dart code runs as expected in debug build, not with AOT.

This issue only affects plugins that select between platform code at build time using pubspec.yaml constructs.

One problematic example:
https://github.com/flutter/plugins/blob/main/packages/path_provider/path_provider/pubspec.yaml#L20
in conjunction with 
https://github.com/flutter/plugins/blob/main/packages/path_provider/path_provider_linux/pubspec.yaml#L16

Stuart's proposal is to implement an environmental variable used at flutter build time to select platform.

https://github.com/flutter/flutter/issues/103660

rzr:
Is this related to this issue i am observing in kirkstone branch: "VM snapshot invalid and could not be inferred from settings."

Same as previously reported issue  https://github.com/jwinarske/meta-flutter/issues/36

jwinarske:
@rzr unrelated.  The Dart VM version in Flutter SDK used to build a Flutter application must match Dart VM in engine used.  It's a 1:1 relationship.  You can send arg to VM via command line to skip this check for development scenarios.  See the Dart VM arg list in source.

jwinarske:
I have figured this one out.  Staging a fix now.

jwinarske:
https://github.com/meta-flutter/meta-flutter/pull/344

matsch82:
This issue is still present for me with Flutter 3.16.7 from https://github.com/meta-flutter/meta-flutter/commit/8e0d313901b7da4d1a2ee434075c34aeb8e35439
and it does not always occur.  VideoPlayer is running fine with the Plugin Implementation from [flutter-elinux-plugins](https://github.com/sony/flutter-elinux-plugins) , while Camera Plugin shows the bug.

The elinux platform plugin is simply not loaded in release builds of the camera demo app.

I try to update to lates kirkstone version, lets see if it is fixed there.

PR #123, opened 2022-06-22
c/flutter-app.bbclass: Add symlink to libapp.so

rzr:
This fix application launching in release mode,
ivi-homescreen is looking for lib (along kernel_blob.bin) in assets dir
(ie: --a=/usr/share/flutter/gallery/flutter_assets/ )

Observed issue was:

    [ERROR:flutter/runtime/dart_vm_data.cc(18)] \
    VM snapshot invalid and could not be inferred from settings.
    [ERROR:flutter/runtime/dart_vm.cc(269)] \
    Could not set up VM data to bootstrap the VM from.
    [ERROR:flutter/runtime/dart_vm_lifecycle.cc(84)] \
    Could not create Dart VM instance.
    [FATAL:flutter/shell/common/shell.cc(143)] \
    Check failed: vm. Must be able to initialize the VM.
    Aborted

Relate-to: https://booting.oniroproject.org/rzr/meta-oniro-blueprint-flutter/-/issues/3#note_24362
Relate-to: https://github.com/meta-flutter/meta-flutter/issues/115#issuecomment-1160263129
Signed-off-by: Philippe Coval <philippe.coval.ext@huawei.com>

### DIFF
--- a/classes/flutter-app.bbclass
+++ b/classes/flutter-app.bbclass
@@ -221,7 +221,7 @@ do_compile() {
     fi
 }
 
-INSANE_SKIP:${PN} += " ldflags libdir"
+INSANE_SKIP:${PN} += " dev-so ldflags libdir"
 SOLIBS = ".so"
 FILES:SOLIBSDEV = ""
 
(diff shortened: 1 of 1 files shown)


jwinarske:
@rzr Heads up.  I introduced a breaking change in ivi-homescreen today.

--a is now for setting Engine accessible features
--b replaces prior usage of --a.

In the next couple of days I will be implementing the following folder structure (based on `flutter build linux` output):

```
            <flutter app install folder>
           ├── data
           │   └── flutter_assets
           └── lib
               └── libapp.so
```

This supports x86_64 homescreen host dev scenarios, as well as standardizing on `flutter build linux` build output format.  A script is in development to support dev use which create this folder structure for an app, as well as download all the required artifacts.  It will be checked into meta-flutter.  This makes it easy to use homescreen for desktop, switch between engine flavors, and run Vulkan backend.

Homescreen will stat <app bundle dir>/data/icuctl.dat before defaulting to load /usr/share/flutter/icuctl.dat.

The flutter app install will reflect the above folder layout.

PR #135, opened 2022-07-22
Fix flutter-pi patch application

bojidartonchev:
Regeneration of the patch was needed after [these](https://github.com/ardera/flutter-pi/commit/04ef38a4e1581deec11ed801b89eb551ba881465) changes.


jwinarske:
Thanks for the patch @bojidartonchev

Issue #209, opened 2022-10-26
Remove path patches for flutter-pi


ardera:
I just merged https://github.com/ardera/flutter-pi/pull/290, you can now enable the meta-flutter filesystem layout in flutter-pi by configuring with `cmake ... -DFILESYSTEM_LAYOUT=meta-flutter ...`.

So https://github.com/meta-flutter/meta-flutter/blob/kirkstone/recipes-graphics/flutter-pi/files/0001-path-updates.patch is no longer necessary (and will probably also fail to apply now)

jwinarske:
@ardera That's great, thanks.  I'll take a look in a few days.

ardera:
Closing because #210 #211 #212 were merged

PR #210, opened 2022-10-26
Use flutter-pi filesystem layout option (kirkstone)

ardera:
Modifies the flutter-pi recipe:
- to configure flutter-pi using `-DFILESYSTEM_LAYOUT=meta-flutter` (lookup rules [here](https://github.com/ardera/flutter-pi/blob/0aec3cc5c9449c4227032af9c853a5047acc5491/src/filesystem_layout.c#L263-L277))
- no longer pin flutter-pi to a specific revision
  - bc there's no longer a patch that can stop working when the flutter-pi is updated
  - though if you pinned it for different reasons, we can also pin it to the current master commit instead
  - (or set up some kind of roller, that opens PRs when flutter-pi updates, like flutter has)
- no longer apply the old path patch
- delete old path patch


ardera:
resolves #209

jwinarske:
@ardera this will need a patch for honister, dunfell, and kirkstone.  Each canary build needs to pass (linux-dummy), also a 32-bit build.

ardera:
@jwinarske I opened PRs for dunfell & honister now as well

jwinarske:
Thanks Hannes!

ardera:
@jwinarske np!

Now that the flutter-pi ref is unpinned again in meta-flutter, should we set up some sort of testing infra for flutter-pi with meta-flutter?

Recently had a build breakage on debian buster, would be cool if we can ensure that doesn't happen for dunfell/kirkstone

(ik you test the flutter-pi build in your CI, but that's only triggered on meta-flutter, not flutter-pi updates, right?)

ardera:
@jwinarske okay, so does AGLs LAVA run on your labs board farm?

> When I implemented the bundle FS patch for flutter-pi a while back, I augmented CMake for Ubuntu host builds. In general it would be good that have this option in flutter-pi. It's easy to switch to a DRM capable terminal. It could also be used for running headless CI tests.

I'm absolutely open to any improvements, though not 100% sure I understand you correctly, you can already build flutter-pi on a normal desktop ubuntu machine right now
Running that on there if a desktop environment is running won't work of course, though I could implement a mock display mechanism so you can still run headless tests.

> It's easy to switch to a DRM capable terminal

do you mean a VT? And then just switch to another VT using `Ctrl+Shift+F2` or something? That'd indeed be useful to test flutter-pi without terminating your user session. I think nowadays that session switching should be done using systemd-logind but not really sure how that works.

jwinarske:
@ardera AGL has a board farm which meta-flutter artifacts are tested on each build.

Yes, correct.  Close all apps, switch to a VT via `Ctrl+Alt+F3`, login, and issue `sudo systemctl stop gdm3`.  This relinquishes atomic access to DRM.

ardera:
@jwinarske Okay, just created a flutter-pi issue to track this: https://github.com/ardera/flutter-pi/issues/300

PR #229, opened 2022-12-14
Static revisions

witekio-kwe:
this is draft PR to show how statically created assets (created by a new script) could avoid the need for network calls in flutter-version.inc


witekio-kwe:
Updates to these files could be created by an automated job (github actions for instance) - no need to do it manually

PR #264, opened 2023-03-15
flutter-engine: allow local version pinning

witekio-kwe:
currently the local patches in flutter-engine do only fit for the proposed default version of FLUTTER_SDK_TAG.
If one would want to locally pin the version to something different one or more patches wouldn’t apply.
Fix this by weak defining a new variable called FLUTTER_ENGINE_PATCHES which can be easily overridden in a local setup or a layer on top of meta-flutter to have the right set of patches stored there


jwinarske:
Please sign your commit
```
git ammend -s`
git push --force
```

PR #265, opened 2023-03-15
gn: add more info on unpack failure

witekio-kwe:
as just the return code is everything but helpful to debug any issues when unpacking.
Use check_output for this purpose and error out with more details.

Also remove the pointless return statement


jwinarske:
LGTM.  Just need you to sign the commit

PR #269, opened 2023-03-20
gn fetch: premirror compatibility

camaronut:
Hi meta-flutter folks,

This PR reflects what we did to accomplish pre-mirroring a download of flutter-engine.

It's sort of a mixed bag.  As we can't make the tar file reproducible without breaking the build, we are stuck with updating our mirror with a new tar file every time we make a release.  I left in some of the not-so-impacting reproducibility functionality, but removed some of the more drastic items (the .cipd/tagcache and PNG file removals are two examples).

What follows below is the commit message.

An attempt was successfully made at creating a fully reproducible tar file for creating a premirror.  Unfortunately, testing showed that the build eventually called git rev-parse, which then failed due to stripping the .git directories out of the tar file.

So a different tack was taken: to accept that similar to git mirrors, gn mirrors will always change and not be reproducible.  Code was primarily borrowed from the git fetcher, but also the repo fetcher as it has some similar characteristics.  The supports_checksum() function was added and returns False now.


jwinarske:
@camaronut  This is a good start.

I'll investigate options for stripping the source tree of non-deterministic timestamps.  I'm thinking it's possible.

jwinarske:
I started investigation here:
https://github.com/meta-flutter/meta-flutter/issues/271

Looks like there is an obfuscated dependency on .git

jwinarske:
Thanks for this @camaronut, and apologies it sat for so long.

The premirror half has now been reimplemented in #776 against the current fetcher, which has diverged quite a bit since this was written — `lib/gn.py` has gained `gclient_config`, `EXTRA_GN_SYNC`, and a config hash in the tarball name, so rebasing a dunfell-era diff would have been mostly conflict resolution. The structure in #776 follows the approach you took here: `supports_checksum() → False`, `try_premirror()` modelled on the git fetcher, and telling bitbake about the mirrorable tarball so `PREMIRRORS`/`MIRRORS` are matched against the artifact name rather than the `gn://` url.

It's verified end to end: with a local mirror seeded and both `BB_NO_NETWORK=1` and `BB_FETCH_PREMIRRORONLY=1`, `do_fetch` succeeds, where the identical configuration previously fell through to `Trying Upstream` and failed with `NetworkAccess`.

The reproducibility half is deliberately not carried over, and your own "mixed bag" framing turned out to be right for reasons that have only sharpened since:

- The premise still doesn't hold — the tarball can't be reproducible while the `.git` directories have to stay for `git rev-parse` (#271).
- The `.packages` and `package_config.json` files the seds rewrote **no longer exist** in a current engine tree — a 3.47.1 checkout has zero of each, and `.packages` has been deprecated in Dart for years.
- `SOURCE_DATE_EPOCH` isn't meaningful during `do_fetch` anyway: `do_deploy_source_date_epoch` is `addtask ... after do_patch`, so `urldata_init` would only ever see the fallback value.

Closing in favour of #776, which credits this PR. The `.git` dependency remains open in #271 — if that's ever resolved, the reproducible-tar work here becomes viable again and would be worth revisiting.

jwinarske:
Correction to my previous comment, @camaronut — I closed this in favour of #776, and **#776 has itself now been closed as a no-op.**

Testing it properly (a control run I should have done before opening it) showed that premirrors already work for gn recipes on master. `uri_replace()` derives the mirror filename from `basename(ud.localpath)` whenever the fetcher's `supports_checksum()` is True, which for a plain tarball in `DL_DIR` it is by default. Declaring `mirrortarballs` adds nothing, and the `supports_checksum() → False` that #776 paired it with would have *broken* that path on its own.

So the accurate statement is not "superseded by #776" but: **the premirror problem this PR set out to solve no longer exists** in the current fetcher, which has moved on considerably since 2023. The mechanism works; what was missing was only the knowledge of how to configure it — use the generic `.*://.*` pattern, and place the tarball under `<mirror>/<url-path>/`. That's going in the README.

Your `supports_checksum() → False` here was reasonable in the fetcher of the day, and the `try_premirror()` you modelled on the git fetcher is still the right shape if the fetcher ever grows a separate clone directory to mirror alongside the tarball.

The reproducibility half remains blocked on the `.git` dependency in #271, as covered above. Staying closed, but for a better reason than the one I gave.

PR #257, opened 2023-03-05
Local Resolution of Flutter SDK and Engine versions

jwinarske:
Avoids use of network calls to determine Flutter Engine Commit

PR #276, opened 2023-04-05
Local Resolution of Flutter SDK and Engine versions (#257)

WAKazuyoshiAkiyama:
This is the backport of https://github.com/meta-flutter/meta-flutter/pull/257.


jwinarske:
@WAKazuyoshiAkiyama Note all dunfell based BSPs that support the override specifier introduced in kirkstone (it was back ported to dunfell), will work as-is with the kirkstone branch.

PR #279, opened 2023-04-18
Add wayland to DEPENDS even without vulkan

### LINKED ISSUE #278: wayland-client.h file not found
I am building flutter-pi with kirkstone branch and by default, it is using FLUTTER_SDK_TAG = "3.7.7". While building flutter-pi, I see flutter-engine fail with the following error:
| In file included from ../../flutter/flutter_vma/flutter_vma.cc:19:
| In file included from ../../flutter/flutter_vma/flutter_vma.h:7:
| In file included from ../../third_party/vulkan_memory_allocator/include/vk_mem_alloc.h:130:
| ../../third_party/vulkan-deps/vulkan-headers/src/include/vulkan/vulkan.h:41:10: fatal error: 'wayland-client.h' file not found
| #include <wayland-client.h>
|          ^~~~~~~~~~~~~~~~~~
| 1 error generated.

I am wondering if there are hard dependencies on Wayland.

FranzForstmayr:
When not using vulkan wayland is still needed to build `flutter-engine`.

This should fix #278

### DIFF
--- a/recipes-graphics/flutter-engine/flutter-engine_git.bb
+++ b/recipes-graphics/flutter-engine/flutter-engine_git.bb
@@ -18,6 +18,7 @@ DEPENDS += "\
     compiler-rt \
     libcxx \
     zip-native \
+    ${@bb.utils.contains('DISTRO_FEATURES', 'wayland', 'wayland', '', d)} \
     "
 
 FLUTTER_ENGINE_PATCHES ?= "\


jwinarske:
Haven't had time to test, but I believe the Wayland dependency is related to impeller option.  If this is the case I would prefer the dependency added to the package config flag responsible

FranzForstmayr:
Impeller is only activated by default  if Vulkan is in DISTRO_FEATURES, isn't it?

PR #280, opened 2023-04-21
Fix flutter gallery

Markus43:
Update to the flutter gallery revision recommended by @ardera in the flutter-pi repository.

With the old SRCREV, flutter gallery does not build for flutter 3.7.x because of some issues with translation files.

This is the only recipe from flutter-test-apps that currently cannot be built with flutter 3.7.x .


ardera:
I have a script btw to find the latest working flutter gallery version btw: https://gist.github.com/ardera/4d9807e1cb05b4e26be2133668197768

jwinarske:
@Markus43 hi Markus.  Thanks for the contribution.  Currently I have a backlog of things to do in meta-flutter.  This is one.

@ardera I'll take a look at your script when I'm back in the office next week.  One approach is to test the SDK version range in the pubspec.yaml.  it is a generally annoying problem for maintenance.  I think there is a case to support auto rolling of recipes.  The flutter watch repro already monitors stable channel changes.  It could be used to stage an auto update.  An auto update would still require some form of test pass.  Which leads to some kind of headless screen grab scenario for evaluating app health.  I would love to avoid the busy work, and come up with some reliable automation

jwinarske:
@ardera I added the pattern as an option to `workspace-automation`:
https://github.com/meta-flutter/workspace-automation/pull/2

Usage pattern:
```
cd app/gallery
$FLUTTER_WORKSPACE/flutter_workspace.py --find-working-commit
```

PR #284, opened 2023-05-09
flutter-engine: allow install of older version

kwavnet:
some of the tools installed are only available in 3.7+ releases of the engine and some of them are bound to impeller-vulkan configuration.
Allow installation on older releases by ignoring missing files (for the ones that are not bound to a PACKAGECONFIG) or bind them to the respective configuration.


jwinarske:
LGTM.  Thanks!

PR #296, opened 2023-06-19
depot-tools-native: fix INSANE_SKIP

kwavnet:
as depot-tools already is a native only package, there isn't any use in defining BBCLASSEXTEND.
Futhermore native recipes do not have any kind of packaging, rendering the assignments of FILES:${PN}-dev and INSANE_SKIP:${PN}-dev not applicable.
Remove any packaging related setting and just use INSANE_SKIP:${PN}. Fixes: ninja-linux64 from depot-tools-native was already stripped, this will prevent future debugging!

### DIFF
--- a/recipes-devtools/depot-tools/depot-tools-native_git.bb
+++ b/recipes-devtools/depot-tools/depot-tools-native_git.bb
@@ -33,10 +33,6 @@ do_install() {
     cp -rTv ${S}/. ${D}${datadir}/depot_tools
 }
 
-FILES:${PN}-dev = "${datadir}/depot_tools/*"
-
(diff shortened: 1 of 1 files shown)


jwinarske:
@kwavnet How was this tested?  SDK and OS image?

kwavnet:
@jwinarske OSImage - popped up in the build log

jwinarske:
It gets installed in SDK build

kwavnet:
and it has a dependency at flutter-engine - I don't remember where it popped up as we are building both OS images and SDK, nevertheless we got this INSANE_SKIP warning, which made we look more closer at this recipe

jwinarske:
got around to running a local build to test

PR #306, opened 2023-07-17
Bump depot_tools revision

rogerzanoni:
This fixes the following issue while bootstrapping depot_tools as a dependency of CEF on AGL:

Could not find a version that satisfies the requirement setuptools>=40.3.0 (from google_api_core==1.31.5->-r /tmp/vpython_bootstrap882189871/requirements.txt (line 22)) (from versions:) No matching distribution found for setuptools>=40.3.0 (from google_api_core==1.31.5->-r /tmp/vpython_bootstrap882189871/requirements.txt (line 22))

### DIFF
--- a/recipes-devtools/depot-tools/depot-tools-native_git.bb
+++ b/recipes-devtools/depot-tools/depot-tools-native_git.bb
@@ -8,7 +8,7 @@ LIC_FILES_CHKSUM = "file://LICENSE;md5=c2c05f9bdd5fc0b458037c2d1fb8d95e"
 SRC_URI = "git://chromium.googlesource.com/chromium/tools/depot_tools;protocol=https;branch=main \
            file://0001-disable-ninjalog_upload.patch"
 
-SRCREV = "3a56ba9d9c9d22bc78e24f96a9096247d53649f8"
+SRCREV = "64b61755572b1d03b3a43f1a29b617dcae3a3fe0"
 
 S = "${WORKDIR}/git"
 


jwinarske:
@rogerzanoni Thanks Roger!

jwinarske:
@rogerzanoni also for reference, what build environment are you seeing this issue?  I have not seen this reported before, so I am very curious.

jwinarske:
I'm hitting a build error with this commit:
```
ModuleNotFoundError: No module named 'httplib2'
```
Build environment and how did you test?

jwinarske:
I am reverting this until I have more info from you.

rogerzanoni:
Hi Joel, sorry for the lack of context, this was for a new recipe that needs depot_tools: https://gerrit.automotivelinux.org/gerrit/c/AGL/meta-agl-demo/+/29036

I was adding another depot_tools recipe (https://gerrit.automotivelinux.org/gerrit/c/AGL/meta-agl-demo/+/29035) but we decided to keep only the one from meta-flutter. I can submit another patch, making the SRCREV default to the one prior to my change and make it configurable so I can set it to one revision that works for this recipe as well

jwinarske:
I see.  Yeah whatever change needs to happen in this layer it needs thorough testing.  The provided source commit is not happy with flutter-engine builds.  What I do is run python2 from the depot_tools cipd path, and then prevent it from requesting update via network (LTS).  When I run with your commit, it fails badly.  I think it's a matter of sorting out the new paths.  Also the python2 requirement has gone away.

If you need an override, you can use a bbappend.  Only that would break the support for the flutter-engine build.

I think this commit is possible, but it will require some time to sort out the dependencies.  Perhaps in a couple of weeks?

rogerzanoni:
We ended up merging other recipe (https://gerrit.automotivelinux.org/gerrit/c/AGL/meta-agl-demo/+/29035) that uses the revision supported by CEF, but we can get back to working on this later, perhaps I can use a bbapend file and only change change the srcrev if we are buiding for CEF to ensure it doesn't break the flutter builds

PR #317, opened 2023-08-23
all: address do_compile proxy issues

StaticRocket:
The compile phase of certain packages depends on a loopback websocket connection. If a proxy is exported for this without a corresponding no_proxy config containing localhost, ::1, and 127.0.0.1, it will result in a connection issue and error out.

If we're exporting most of the proxy variables manually now then lets just use the export_proxy command from bitbake directly were proxies are needed.

### DIFF
--- a/classes/gn-for-flutter.bbclass
+++ b/classes/gn-for-flutter.bbclass
@@ -1,8 +1,10 @@
 # Copyright (c) 2020-2022 Woven Alpha, Inc
 
 python () {
+    from bb.utils import export_proxies
     import gn
     bb.fetch2.methods.append(gn.GN())
+    export_proxies(d)
 }
 
 DEPENDS += " \
@@ -31,10 +33,6 @@ do_fetch[depends] += " \
 
 do_configure[network] = "1"
 do_configure:prepend() {
-    export http_proxy=${http_proxy}
-    export https_proxy=${https_proxy}
-    export HTTP_PROXY=${HTTP_PROXY}
-    export HTTPS_PROXY=${HTTPS_PROXY}
     export PATH=${DEPOT_TOOLS}:${DEPOT_TOOLS}/${PYTHON2_PATH}:${PATH}
     export DEPOT_TOOLS_UPDATE=0
     export GCLIENT_PY3=0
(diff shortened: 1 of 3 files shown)


StaticRocket:
Ugh. I was trying to be slick but the do_compile stage throws out any previously exported env vars unless the variable attribute explicitly has export set. One moment.

StaticRocket:
Let me just go straight to the source of the headache.

StaticRocket:
https://lists.openembedded.org/g/bitbake-devel/message/14959

jwinarske:
I like the idea, and it looks like a more elegant solution

StaticRocket:
@jwinarske , thanks, please respond to the mailing list with that affirmation if you have a chance. It's gotten a little off topic with the discussion of some other tooling that may be abusing the network task flag...

If this layer has a good example of the network flag being used correctly please post it there. (I'm not too familiar with this layer in particular.)

https://lists.openembedded.org/g/bitbake-devel/message/14968

jwinarske:
@StaticRocket the network flag being used correctly is interesting.  As in enabling the network availability per task?

GN vs Yocto is something this layer has been working with.  Currently we have a fairly efficient gn fetcher.  Should work with any GN based project.  I recently tested a script to generate a Yocto recipe from a gn project.

Flutter/Dart have funny manifests that pull from network during build.  So more work is needed to support LTS for flutter apps.

StaticRocket:
Right. I know how frustrating some of these mechanisms are but to Yocto and Bitbake any network access outside of do_fetch is a problem because now it introduces independent variables into what should be a well defined, purely state driven system.

I myself have a few recipes that break this rule but that's why I was curious if you had some tasks explicitly defined that were outside of the scope of do_fetch that required networking. I've been cheating (and so have the meta-openembedded guys with influx and other go based recipes) so I don't actually have anything on hand.

jwinarske:
I see.  It's in a couple of spots.  A grep will list them all.
Getting Flutter apps to not download critical artifacts during build will be a challenge to solve in an agnostic+scalable way.  I think it's doable though.

That said I don't think this layer would work without being able to override the default network behavior per task.

PR #319, opened 2023-08-29
Fix compiling against 3.13.x

aduskett:
https://github.com/flutter/flutter/issues/127606 has been fixed for quite some time and is no longer needed. In addition, the inclusion of https://dart.googlesource.com/tools.git@545d7e1c73ce21b8c91f638021f9d487d324a501 causes 3.13.x to fail with the following error:

output: Error: Couldn't resolve the package 'linter' in 'package:linter/src/rules.dart'.
analytics_manager.dart:223:23: Error: The method 'send' isn't defined for the class 'Analytics'.

Remove the inclusion of unified_analytics.


aduskett:
No problem for me. I noticed it when updating flutter with my Buildroot patch series. Feel free to deny it.

PR #336, opened 2023-10-16
Roll homescreen and flutter-auto

jwinarske:
-rename with-logging class to verbose-logs

-ivi-homescreen
 * CI Updates
 * Update Packaging Info
 * UT missing method
 * AGL cherry picks

-flutter-auto
 * remove from clang dynamic layer.  Only built with GCC now.
 * rebased to ivi-homescreen main
 * maintain flutter-auto name
 * aside from name all logic is at parity with ivi-homescreen


jwinarske:
cc @g-scott-murray

PR #338, opened 2023-10-23
ivi-homescreen_git.bb: remove duplicated rdepends

emanuel-th:
Rdepends to flutter-engine has been added twice to the recipe over time. See line 24.


jwinarske:
@emanuel-th Hi.  Thanks for the PR.  Looks like I have it duplicated in flutter-auto as well.  Can you add change for that as well?

emanuel-th:
@jwinarske I added the change for flutter-auto.

jwinarske:
@emanuel-th thanks!

PR #351, opened 2023-11-09
Remove pyyaml dependency

g-scott-murray:
Replace pyyaml module usage in app bbclass code with re based parsing of the app name from pubspec.yaml.  This avoids needing either pyyaml on the build host or mixing the host and -native Pythons as was being done. Also, a couple of the app name errors have been bumped to bb.fatal from bb.error to end building immediately, as do_compile is going to fail in those situations, anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+Nov 9, 2023
+1. Replace pyyaml use with re to avoid mixing host and Yocto -native
+   Python bits.
+2. Changed a couple of pubspec.yaml name errors to bb.fatal to stop
+   build immediately, as do_compile will fail.
+
 Nov 8, 2023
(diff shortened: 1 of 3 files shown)


dl9pf:
+1 from my side.

jwinarske:
This file and references to it's functions would need to be removed as well:
```
conf/include/app-utils.inc
```

I think I will transition compilation and pubspec parsing to dart.  It's available and would be quicker to iterate on.

g-scott-murray:
I think I removed all of the unused code, the couple of functions that remain are what is called from do_compile pretty much as previously?  Unless you mean you would like it all hoisted into do_compile until you do further rework?

jwinarske:
oh right.  Just reviewed on my machine, instead of phone...

LTGM

PR #358, opened 2023-11-20
Add a way to disable obfuscation if needed

### LINKED ISSUE #357: Add a way to build without --obfuscate flag
Since obfuscation hides function and class names in compiled Dart code, there are some cases, when it should be avoided. For example when using [runtimeType](https://api.flutter.dev/flutter/dart-core/Object/runtimeType.html).

bojidartonchev:
Fixes #357


jwinarske:
--obfuscate is hard coded right now for a reason.  Without it I don't see vm:entry work correctly on projects with "remote" packages.  It's not needed for "local" package unit test cases.  I have an open issue on flutter.  I need to simplify the repro case for Stuart.

Once the issue is addressed I will move --obfuscate to append to default value of `app_gen_snapshot_flags`.

Issue #360, opened 2023-11-22
Random error in flutter-engine compile`: scene_flatbuffers.h file not found


kwavnet:
```
| In file included from ../../flutter/display_list/benchmarking/dl_complexity.cc:6:
| In file included from ../../flutter/display_list/benchmarking/dl_complexity_gl.h:8:
| In file included from ../../flutter/display_list/benchmarking/dl_complexity_helper.h:10:
| In file included from ../../flutter/display_list/dl_op_receiver.h:10:
| In file included from ../../flutter/display_list/dl_canvas.h:9:
| In file included from ../../flutter/display_list/dl_paint.h:12:
| In file included from ../../flutter/display_list/effects/dl_color_source.h:25:
| In file included from ../../flutter/impeller/scene/node.h:18:
| ../../flutter/impeller/scene/animation/animation.h:18:10: fatal error: 'impeller/scene/importer/scene_flatbuffers.h' file not found
| #include "impeller/scene/importer/scene_flatbuffers.h"
|          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
| 1 error generated.
| [8078/9224] CXX clang_x64/obj/third_party/vulkan-deps/glslang/src/SPIRV/glslang_sources.GlslangToSpv.o
| [8079/9224] CXX obj/flutter/display_list/display_list.dl_builder.o
| FAILED: obj/flutter/display_list/display_list.dl_builder.o
```

We get this error randomly in flutter-engine compile.
Using poky kirkstone 4.0.14 (in before we never have seen that).
In most cases it works absolutely fine, but like in every 15-20th job this one errors out with above message.

Is that a know issue, like a race between fetching and building?
I admit I'm a bit lost where to look first here

jwinarske:
@kwavnet 
It might be a bug in the engine build itself.  If that file is being generated, it very well could be a race condition.

You try removing `impeller-3d` from the package config variable.  Then see if the problem goes away.  If that addresses it I'll remove it in the kirkstone branch.

kwavnet:
@jwinarske thanks, will try that

kwavnet:
@jwinarske it seems like it did the trick - builds are stable now with `impeller-3d`

jwinarske:
@kwavnet with our without?  The default is currently enabled by default

kwavnet:
> @kwavnet with our without? The default is currently enabled by default

without ofc, sorry

[...]

PR #365, opened 2023-11-29
Remove impeller-3d from flutter-engine packageconfig

jwinarske:
-reports of engine build failing every 15-20 times with this flag enabled.


jwinarske:
resolves https://github.com/meta-flutter/meta-flutter/issues/360

PR #367, opened 2023-12-07
Support offline builds

onlyxuyang:
Modifiy code to facilitate offline builds. 
Changes include: 
1) re-generate package json after clean in flutter-app.inc 
2) make clean cache as a task in flutter sdk
3) keep necessary files in flutter sdk


jwinarske:
Currently an offline pubcache is stored in DL directory before compilation.  Which prevents needing network access for compile, which is required for LTS support.  See:  https://github.com/meta-flutter/meta-flutter/blob/kirkstone/conf/include/flutter-app.inc#L51-L118

Can you describe what your proposed change affects?  Yocto SDK build, or?

onlyxuyang:
> Currently an offline pubcache is stored in DL directory before compilation. Which prevents needing network access for compile, which is required for LTS support. See: https://github.com/meta-flutter/meta-flutter/blob/kirkstone/conf/include/flutter-app.inc#L51-L118
> 
> Can you describe what your proposed change affects? Yocto SDK build, or?

Currently pubcache cannot make the flutter built successfully without network access, I am not sure current "offline build" is tested under setting 'BB_NO_NETWORK = "1" ', or really tested under a no network environment.  Task network access is enabled by do_<task>[network] = "1" in some places so we can not test "offline build" only by setting "BB_NO_NETWORK"

In my side, I cannot build pass only with the pubcache and I need no network access at all.

I achieve the goal but because no suitable function hook point, I need copy full function in bbappend and modify in my local. It is not well for further maintain, so I submit current patch which not break current meta-flutter status but let I can maintain modification in independent function.

For current patch, 
1) modification in flutter-sdk_git.bb is for a complete cache for sdk build. Current do_compile remove cache before precache which make it must access network.  And remove necessary files for app build in do_install to make app must access network.
2) modification in flutter-app.inc(moved in common.inc in lastest version) is because if a "flutter pub get --offline" not after clean, it seems to access network for something.

Thank you

jwinarske:
There is much more than your PR addresses to enable offline building.  It's a good start and I appreciate your PR.

* AUTOREV
  building without a network connection all cases of ${AUTOREV} fail:  https://github.com/meta-flutter/meta-flutter/pull/377
* do_cleanall - stale pub caches are not helpful when debugging offline support:  https://github.com/meta-flutter/meta-flutter/pull/378
* all instances of `do_compile[network] = "1"` need removal
  - this involves downloading addition deps and adding to flutter-sdk-native; such as material fonts.
* SDK install testing needs to happen, as this change breaks SDK build scenarios:
```
rm -rf ${S}/bin/cache/pkg/sky_engine/
rm -rf ${S}/bin/cache/artifacts/*
```

My approach for testing is to remove the network connection (air gap) on my machine.

jwinarske:
I looked at this further, and your change is acceptable.  Thank you for your PR!

PR #369, opened 2023-12-15
Enable embedded web server building in flutter

Gmatarrubia:
Since flutter can build web app in native javascript code, it could be used to build a embedded web server/app.
I added a flutter-web.bbclass for building an EWS compilation.
This EWS can be powered with nginx or other web engines.

### DIFF
--- a/classes/flutter-web.bbclass
+++ b/classes/flutter-web.bbclass
@@ -1,0 +1,223 @@
+#
+# Gonzalo Matarrubia's effort based on the previous work
+# from Joel Winarske. See flutter-app.inc for credits.
+# Helper class for building a Flutter Web Application.
+# Useful for embbeded web services.
+# Assumes that:
+# - S is defined and points to source directory.
+# - PUBSPEC_APPNAME is defined correctly.  This is the name value from pubspec.yml.
+#
+
+require conf/include/flutter-version.inc
+require conf/include/app-utils.inc
+require conf/include/common.inc
+
+DEPENDS += " \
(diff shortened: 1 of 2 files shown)


Gmatarrubia:
@jwinarske I would appreciate your thoughts on this. I'm currently using this code successfully.

jwinarske:
@Gmatarrubia Hello and thanks for the contribution!
This is great, thanks for taking this on.

Looking at this my first impression is that it would be best to incorporate into the existing flutter-app.inc.  just make it a different platform type.  I have something similar planned for building Linux gtk projects.

Let me make a first pass on this and I will provide some further guidance

Gmatarrubia:
Yes, that was my first though. But I after some tries I turned to this approach for this reasons:

- I don't want to add dependence of flutter-engine. It has a big impact in the building time and it is not needed at all.
- There are a lot of steps in compile task that have to be skipped.
- Perhaps the complexity of flutter-app class increase too much.

Anyhow, If you consider that It would be worth to try, I can try it.

jwinarske:
First pass:  https://github.com/meta-flutter/meta-flutter/pull/373
Validated with gallery web recipe.  Results don't get past loading page on firefox.  Not sure what the required setup there is.  I just loaded it using `file://<web>/index.hml`.  Is there any special setup?

I split the includes apart while maintaining as much common as possible.  This will make it easier when I get around to adding GTK support at some point, as it will be flutter-gtk.bbclass/flutter-gtk.inc.

PR #383, opened 2024-01-18
Update CI container

jwinarske:
-remove runtime dependency on file-selector


jwinarske:
moving to freedesktop for CI as gitlab-runner honors DOCKER_HOST variable.

Issue #387, opened 2024-01-22
Package build stalls and crashes for debug builds


matsch82:
Hi,

i am currently working on a project that uses flutter on a Xilinx device, thus we have to use petalinux and do some patches to get it working. 
Unfortunately. the app debug bulds are not working. It shows a strange behaviour that is probably not related to the patches, but more to the  flutter sdk tooling.

In this example the `flutter-packages-example-video-player`, but it the same for all packages.  Release builds are working fine, but if ` FLUTTER_APP_RUNTIME_MODES = "debug"`  is set, 
flutter app build process crashes after a very long time, and consuming 30+ GB RAM.

```
ERROR: flutter-packages-example-video-player-git-r0 do_compile: failed -6 (cmd was /home/peta/projects/share/petalinux/build/tmp/work/cortexa72-cortexa53-xilinx-linux/flutter-packages-example-video-player/git-r0/git/engine_sdk/sdk/clang_x64/gen_snapshot --deterministic --snapshot_kind=app-aot-elf --elf=libapp.so --strip --obfuscate .dart_tool/flutter_build/*/app.dill):                                                                   
../../third_party/dart/runtime/vm/kernel_loader.cc: 221: error: Trying to load a concatenated dill file at a time where that is not allowed 
version=3.1.5 (stable) (Tue Oct 24 04:57:17 2023 +0000) on "linux_simarm64"
  pid=768640, thread=768640, isolate_group=isolate(0x55f1c35e8390), isolate=(nil)((nil))                                                                                                                                     os=linux, arch=arm64, comp=no, sim=yes                                                                                                                                                                                     isolate_instructions=0, vm_instructions=0                                                                                                                                                                                  fp=7ffeeae0df60, sp=7ffeeae0de28, pc=55f1c30cedec
  pc 0x000055f1c30cedec fp 0x00007ffeeae0df60 /home/peta/projects/share/petalinux/build/tmp/work/cortexa72-cortexa53-xilinx-linux/flutter-packages-example-video-player/git-r0/git/engine_sdk/sdk/clang_x64/gen_snapshot+0x446dec
  pc 0x000055f1c2f3fae4 fp 0x00007ffeeae0e040 /home/peta/projects/share/petalinux/build/tmp/work/cortexa72-cortexa53-xilinx-linux/flutter-packages-example-video-player/git-r0/git/engine_sdk/sdk/clang_x64/gen_snapshot+0x2b7ae4
[...]

PR #388, opened 2024-01-22
Fix attempting AOT compile for debug builds

TeknoVenus:
Invalid if statement meant that the recipe attempted to build AOT snapshot for debug builds.

Potentially fixes https://github.com/meta-flutter/meta-flutter/issues/387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+Jan 22, 2024
+1. Fix debug builds for apps
+
 Jan 19, 2024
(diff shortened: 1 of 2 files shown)


matsch82:
this acutally fixed the broken debug build process!

jwinarske:
@TeknoVenus nice find!  Yes that is incorrect.

Note that there is a standalone script for updating AOT release/profile on a device here which is a direct port of the compile steps:
https://github.com/meta-flutter/workspace-automation/blob/main/create_aot.py
You may find it useful.

PR #480, opened 2024-04-27
Address proxy and host contamination issues

StaticRocket:
Fix another set of host contamination issues and proxy problems.

I think having a common python method to export all of these variables, or a bash shim to inject these variables for any native invocation of flutter or dart may also be useful. I'm not really sure what would be the better solution but having these variables all over the place is not great.

Because dart doesn't even follow the basic XDG standards we have to do something a little drastic to make sure it doesn't affect the host machine. We have to override HOME temporarily so that it can lookup it's config. I really don't like this as:

1. It just looks gross
2. It has the potential to interfere with other commands being executed in that same environment.

Open to suggestions there.

### DIFF
--- a/conf/include/common.inc
+++ b/conf/include/common.inc
@@ -106,6 +106,24 @@ python do_archive_pub_cache() {
     # required for flutter: https://github.com/flutter/flutter/issues/59430
     env['XDG_CONFIG_HOME'] = f'{workdir}'
 
+    http_proxy = d.getVar('http_proxy')
+    if http_proxy != None:
+        env['http_proxy']     = f'{http_proxy}'
+
+    https_proxy = d.getVar('https_proxy')
+    if https_proxy != None:
+        env['https_proxy']    = f'{https_proxy}'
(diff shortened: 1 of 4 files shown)


StaticRocket:
Sorry, this shouldn't have been a draft. This is a viable solution but I was trying to make it more of an RFC for future work.

jwinarske:
@StaticRocket I think a cleanup that implements a common file for each Python and Bash exports is in order.  This would clean up readability.  Anytime I start adding things like this I start refactoring, so makes sense

The next pressing priority is a pubspec_fetcher.  Right now disk space is heavily wasted in the case of many apps (100+).

StaticRocket:
Yeah, a fetcher hooking into the new pubspec archive function would be the next logical step to eliminate some of those network overrides...

jwinarske:
I'll start looking at a pubspec fetcher in a couple of weeks.  Lots on my plate right now.  Focused on v2 work.

PR #482, opened 2024-05-02
PUB_CACHE_ARCHIVE incorrectly used SRCREV instead of SRCPV

blindmatrix:
When using SRCREV = "${AUTOREV}" in a recipe SRCREV will resolve to "AUTOINC" without any revision details causing the cache archive filename to always exist after first run, even when app source is updated. SRCPV will instead resolve to something like "AUTOINC+ea5ee74e60" causing new filenames when source revision change.

It should be merged to other branches too if you accept this solution.


jwinarske:
cherry picked to `dunfell` and `scarthgap`.  Thanks!

PR #487, opened 2024-05-10
Add fetcher for pub_cache

SeungkyunKim:
Add new fetcher for flutter pub cache. This fetcher is simillar kind of npmsw, rust or go.
Due to the lack of full dependencies in the pubspec.yaml file, the pubspec.lock file is used.
The reason why the pubspec.yaml file only is insufficient for getting full dependencies is as follows:
  For packages specified as git or path types, it is impossible to
  know the related dependencies until all those packages are fully downloaded.

Add new fetch method Pub is added to handle pubspec.lock file.

SRC_URI += "pub://${THISDIR}/pubspec.lock"


SeungkyunKim:
Dear developers, I have implemented a new pub-cache fetcher feature based on the pubspec.lock file.
This fetcher allows you to build the app offline without using the previouse do_archive_pub_cache and do_restore_pub_cache steps.

By using pubspec.lock, we can ensure that the build results of the app are always consistent due to the pacakge version range.
Additionally, downloaded package files are located under download directory, it is possible to share common package files among apps. All steps are running in do_fetch, do_unpack, thus it can follow the Yocto convention and prevents the use of the network in stages other than fetch.

I would be glad to discuss and share opinions about the code.
Thanks

-B.R. Seungkyun Kim

jwinarske:
@SeungkyunKim hello 👋 and thank you for your contribution 🙂 

I'll take a look at this early next week (US).

We were discussing the plan for this yesterday in the AGL production ready meeting.  Justin from ICS has started work on this.  The key areas that were being discussed were license list support, SPDX, and recipe generation similar to the Yocto script for Rust.  Yocto Rust support does not yet fully support SPDX.  One observation is that each downloaded pub package contains an SPDX file; needs to be confirmed that this is universal.  There was a dialog to start with the Garmin Yocto Dev on best approach to incorporate SPDX for the pub fetcher pattern.

Your collaboration is appreciated!

SeungkyunKim:
Thank you for your quick response. :smile:

jwinarske:
@SeungkyunKim

Initial thoughts:
1.  The design goal for pub spec fetching involves replicating Yocto Rust patterns with the addition of SPDX support:
* Recipe becomes origin of truth
* license listing for each pub package
* hash for each pubspec package
* local git package support
* SPDX support; there is an active dialog with Yocto devs on this topic.
2. The pub spec fetcher is meant to replace the pub cache archive task.  Meaning it will be removed.
3. Licenses are be available in pub.dev fetches in form of SPDX file.  There needs to be a license list in the recipe for each pub.dev package.

Top level goals:
* Software Bill of Materials compliance (SPDX)
* Build reproducibility / Long Term Support - no network access is allowed during patch, configure, or compilation.  Only fetch.

jwinarske:
@SeungkyunKim 

> I believe that specifying every dependent package in a Rust-style will cause significant inconvenience,
> In case of flutter-gallery, it has roughly 100 packages to add on recipe.

Yes I understand your concern.  I have some Rust recipes with 800+ packages.  While that seems like a lot, I have tooling that auto-generates the recipe.  Tooling to generate pubspec based recipes is a requirement, and is being built.  It's simply a matter of running a script on the app.  This script will be used to auto-roll the 100+ the recipes (CI auto roller), as well as be used to create a one off recipe.  The same pattern exists for Yocto Rust.  The consensus at AGL is to follow Yocto upstream patterns.  My plan is to upstream this layer to the Yocto project once LTS and auto-rolling is nearly hands free.

What is a significant inconvenience is auto rolling 100+ apps using a lockfile.  The issue I hit here is that the lock file is relative to a particular SDK.  This same problem happens with Rust.  If you delete the lockfile, and rebuild it with newer SDK, versions may and likely change.  I am treating the lock file as a transient file, which I see as being relative to the target SDK version.  In order to best handle this, I believe the recipe needs to be the origin of truth, which upon creation is relative to the SDK version.

Limiting to use of lock file makes maintenance for 100+ recipes much more time consuming.  It may work fine for a couple, but not many.

Having a checksum on the package detects if the package has changed (I have seen Rive change packages without numbering).  Not having upstream package change detection prevents LTS.  Licensing falls in the same bucket.

I haven't had time to study the pubspec tooling, but I would not want to pull all the packages if they were not going to be used for a given configuration.  For instance Fuschia packages would never be used by Yocto.

My primary driver for a pubspec fetcher aside from LTS is to save disc space.  Building 100+ apps across many SDK starts to eats disc space quickly.  Making recipes relative to a SDK version improves disc space usage.

jwinarske:
@SeungkyunKim closing in favor of https://github.com/meta-flutter/meta-flutter/pull/546

PR #499, opened 2024-06-12
common: add lockfiles for cache ops

kwavnet:
when using a shared DL_DIR across build, e.g. multiple machines with NFS shared DL_DIR, or even multiconfig builds, it could happen that one machine is running do_archive_pub_cache (recreating/updating the cache file), while another runs do_restore_pub_cache. This creates a race condition, where randomly do_restore_pub_cache would fail wil an UnpackError.
To work around this issue, use a shared lock file for the cache artifact across all tasks using it.


jwinarske:
LGTM

PR #500, opened 2024-06-12
flutter-engine: add FLUTTER_ENGINE_STRIP

kwavnet:
which strips the engine lib in a configurable way. By default only release profile is stripped.

It will save ~250MB of disk space when used on the target


kwavnet:
Tested on various arm64, qemu and intel platforms (with release profile being stripped) over the past month, without noticing any issues

jwinarske:
hmm.  I would expect Yocto to do the stripping

kwavnet:
Same here, but the lib so is unstripped being a ~280MB blob

PR #570, opened 2024-09-14
retain partial sync on fetch error

amfern:
If an error occurs during the gclient sync, we aim to retain the partial sync directory.
This helps prevent the need to re-fetch a large amount of source code only to encounter the same network issue again.
The sync directory can grow up to 14GB.
However, it will still be deleted when running bitbake -c cleanall.

To test the change, you can start a clean build of the `flutter-engine`. And before the do_fetch has finished, disconnect the network.(via control panel or any other means).
The do_fetch would fail, however the partial download would be retained in  `${DL_DIR}/gn`, so restarting the build with the network should continue fetching dependencies.

### DIFF
--- a/lib/gn.py
+++ b/lib/gn.py
@@ -64,6 +64,7 @@ class GN(FetchMethod):
         ud.syncpath = os.path.join(gndir, ud.syncdir)
         ud.localfile = ud.syncdir + "-" + srcrev + ".tar.bz2"
         ud.localpath = os.path.join(gndir, ud.localfile)
+        ud.trying_to_fetch_with_gclient = False
 
         sync_opt = d.getVar("EXTRA_GN_SYNC")
         curl_ca_bundle = d.getVar('CURL_CA_BUNDLE')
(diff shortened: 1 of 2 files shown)


lib/gn.py
```diff
@@ -112,6 +113,7 @@ def download(self, ud, d):
 
         uri = ud.url.split(";")[0]
 
+        ud.trying_to_fetch_with_gclient = True
```
jwinarske, line 116: Shouldn't this be set to False after a successful fetch?

jwinarske (changes requested):

lib/gn.py
amfern: Yeah, it's a good idea to put it back to False.

jwinarske (approved):

jwinarske:
@amfern Looks good, just need some documentation on how to test this so it's maintained moving forward.

amfern:
@jwinarske where would like me to add the documentation?

jwinarske:
Document test strategy in the PR, then add a now somewhere in README.md

WAKazuyoshiAkiyama:
LGTM

jwinarske:
@amfern can you comment how you tested this?  That's all I need.

amfern:
To test the change, you can start a clean build of the flutter-engine. And before the do_fetch has finished, disconnect the network.(via control panel or any other means).
The do_fetch would fail, however the partial download would be retained in ${DL_DIR}/gn, so restarting the build with the network should continue fetching dependencies.

Issue #490, opened 2024-05-15
do_archive_pub_cache also copies parent directories 


jgerdes5:
I found a bug while the function 'do_archive_pub_cache' copies the project, where it will also copy the parent directories. This resulted in completely decimating my available hard drive space and finally stopping, because there was no space left. It happens when using bash as a terminal.

The offender is the following line in /conf/include/common.inc:153

`    cp_cmd = \
        'mkdir -p %s/.project | true; ' \
        'cp -r .* %s/.project/ | true; ' \
        % (pub_cache, pub_cache)
`

This will also copy the directories '.' and '..'. The bug is present while using bash or zsh. Other shells I haven't tested.

Here is the log.do_archive_pub_cache file:

<details>
<summary>log.do_archive_pub_cache</summary>
DEBUG: Executing python function extend_recipe_sysroot
NOTE: Direct dependencies are ['virtual:native:/home/username/poky/meta-flutter/recipes-graphics/flutter-sdk/flutter-sdk_git.bb:do_populate_sysroot', 'virtual:native:/home/username/poky/meta/recipes-extended/pbzip2/pbzip2_1.1.13.bb:do_populate_sysroot']
[...]

PR #573, opened 2024-09-27
common: Do not copy parent directories in archive_pub

julianscheel:
Using `cp -r .*` will also include the parent directory `..`, which causes more files to be copied than needed.
Use -T option instead to just copy the current directory as a whole.


jwinarske:
LGTM

jwinarske (approved):
LGTM

jwinarske:
Resolves https://github.com/meta-flutter/meta-flutter/issues/490

PR #585, opened 2024-10-30
Building flutter-pi depends on libsystemd

resilem:
Building flutter-pi recipe will fail if systemd is not declared in `DEPENDS` and it is built before systemd. This commit addresses the issue by ensuring systemd is built before flutter-pi.

### DIFF
--- a/recipes-graphics/flutter-pi/flutter-pi_git.bb
+++ b/recipes-graphics/flutter-pi/flutter-pi_git.bb
@@ -21,6 +21,7 @@ DEPENDS += "\
     libinput \
     libxkbcommon \
     virtual/egl \
+    systemd \
    "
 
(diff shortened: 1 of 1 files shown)


jwinarske:
@resilem 

It builds as-is when using the following in `local.conf`:
```
DISTRO_FEATURES:append = " systemd"
DISTRO_FEATURES:remove = "sysvinit"
INIT_MANAGER = "systemd"
```

Are you building for a sysvinit system?

resilem:
I am building for systemd system. It will build as is _as long as_ systemd is built before flutter-pi.

jwinarske:
okay that makes sense.  So a race condition.  I'm fine with the change.  I just haven't ever seen it happen.  Maybe it's because I have large core count machines.

PR #610, opened 2024-11-25
flutter-auto: check for systemd DISTRO_FEATURES

kwavnet:
as default packageconfig wdt_systemd pulls in systemd, as a dependency, but that requires systemd in
DISTRO_FEATURES.
Add a check to fix that


jwinarske:
Looks good.  Thanks!

PR #614, opened 2024-12-03
ivi-homescreen: check commercial flag

kwavnet:
for video_player_linux ffmpeg is required,
which requires LICENSE_FLAGS_ACCEPTED
commercial to be set


jwinarske (approved):
LGTM

PR #658, opened 2025-05-31
gn.py: small fixes

shr-project:


jwinarske:
@shr-project Thanks for the contribution!

Issue #411, opened 2024-02-12
pdfium does not build with arm64 host


jwinarske:
This is currently a limitation with the PDFium project itself.  I only references linux-x64 in the DEPS file.

The fetch fails looking for linuxarm64 cip asset.

leojrfs:
I can confirm it fails with:

```
failed to resolve infra/rbe/client/linux-arm64@re_client_version:0.148.0.41b09b51-gomaip (line 8): no such package: infra/rbe/client/linux-arm64
```

rbe binaries don't seem to be available for `linux-arm64`:
- https://issues.chromium.org/issues/40277110

leojrfs:
After investigating further, i found a temporary solution.

#### There are currently two issues with `pdfium`:

1. `'buildtools/reclient'` in `DEPS` sets its package name as `'package': Var('reclient_package') + '${{platform}}'`.
This means that when building on `arm64` hosts the package name will solve to `infra/rbe/client/linux-arm64`, which does not exist:
    - https://chrome-infra-packages.appspot.com/p/infra_internal/rbe/client
The same issue was found in chromium and the solution was to enable this package conditionally for non linux-arm64 hosts:
    - https://issues.chromium.org/issues/40277110
    - https://chromium.googlesource.com/chromium/src/+/d2c1621d8e02c804b67cf1722595754114cb9aa%5E%21/#F1

2. `'buildtools/linux64'` in `DEPS` is hardcoded to `gn/gn/linux-amd64`. This is because the `pdfium` version in the recipe available in this repo (at the time of writting this comment) predates the change that fixes this issue:
    - this repo:
        - https://github.com/meta-flutter/meta-flutter/blob/2e600c6a3fd362de6961fea89af85b22d499f4bd/recipes-graphics/pdfium/pdfium_130.0.6721.0.bb#L25
        - https://pdfium.googlesource.com/pdfium/+/2b675cf15ab4b68bf1ed4e0511ba2479e11f1605/DEPS#248

    - upstream fix:
        - https://pdfium.googlesource.com/pdfium/+/cc0f452e0b93e8ad80c096b8b40b3a736b5e7435%5E%21/DEPS


#### Fixes and workarounds

1. Ideally, upstream would implement a similar fix to the one done in chromium. Until then we can either:
    - fork the repo and just comment the `buildtools/reclient` package declaration in `DEPS`
[...]

PR #670, opened 2025-06-16
pdfium: fix building in arm64 hosts

leojrfs:
'buildtools/reclient' in DEPS sets its package name as 'package': Var('reclient_package') + '${{platform}}'. This means that when building on arm64 hosts the package name will solve to `infra/rbe/client/linux-arm64`, which does not exist:

- https://chrome-infra-packages.appspot.com/p/infra_internal/rbe/client

The same issue was found in chromium and the solution was to enable this package conditionally for non linux-arm64 hosts:

- https://issues.chromium.org/issues/40277110
- https://chromium.googlesource.com/chromium/src/+/d2c1621d8e02c804b67cf1722595754114cb9aa%5E%21/#F1

This PR adds a workaround to solve this issue by using `vars` to replace the binary name and version with one that is available for both arm64 and x86_64 platforms, satisfying the dependencies when `gclient sync` runs.

It will not affect the build as `reclient` is not used.

Check:
- https://github.com/meta-flutter/meta-flutter/issues/411#issuecomment-2955621158

### DIFF
--- a/recipes-graphics/pdfium/pdfium_139.0.7228.0.bb
+++ b/recipes-graphics/pdfium/pdfium_139.0.7228.0.bb
@@ -40,7 +40,18 @@ inherit gn-fetcher pkgconfig
 require conf/include/gn-utils.inc
 
 # For gn.bbclass
-GN_CUSTOM_VARS ?= '{"checkout_configuration": "small"}'
+# Trick `gclient sync`. Download some binary in order to satisfy dependencies.
+# Check reasoning at:
+# - https://github.com/meta-flutter/meta-flutter/issues/411#issuecomment-2955621158
+# Keep this workaround while the following issue is unsolved:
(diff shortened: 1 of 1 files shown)


leojrfs:
> @leojrfs make sure you have run this on your branch:
> 
> ```
> git checkout scarthgap
> git pull
> git checkout <your branch>
> git rebase scarthgap
> git push
> ```
> 
> this should fix the CI jobs

This commit is already on top of this repo's `scarthgap`. It's parent is 1bb5fb542eff3fd592f205cb1c3b24403e0205a9, which at the time of writing this is the head of [`scarthgap`](https://github.com/meta-flutter/meta-flutter/tree/scarthgap).

jwinarske (approved):
LGTM

jwinarske:
It was a container issue on my end.  Good to go.  Thanks again @leojrfs !

PR #681, opened 2025-07-07
pdfium: remove static compilation of libgcc and libg++

leojrfs:
The library should not be built with `-static-libgcc` and `-static-libg++` as the rest of Yocto is built to use dynamic libraries.

Mixing static and dynamic C++ runtimes causes crashes during exception handling. When an exception from `pdfium` crosses the boundary to code using the shared `libstdc++`, the runtime environments clash and the program terminates.

This commit removes the static flags, ensuring `pdfium` uses the same dynamic `libstdc++` as the rest of the system, which fixes the underlying issue.

My guess is that this file was copied from the `musl-static` build of the [`pdfium-binaries`](https://github.com/bblanchon/pdfium-binaries project) repo, where those static flags apply, but removing them here was forgotten.

### DIFF
--- a/recipes-graphics/pdfium/files/toolchain.gn.in
+++ b/recipes-graphics/pdfium/files/toolchain.gn.in
@@ -12,7 +12,7 @@ gcc_toolchain("@GN_TARGET_ARCH_NAME@") {
   ld = cxx
 
   @EXTRA_CXXFLAGS@
-  extra_ldflags = "-static-libgcc -static-libstdc++ @LDFLAGS@"
+  extra_ldflags = "@LDFLAGS@"
 
   toolchain_args = {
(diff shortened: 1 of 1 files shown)


leojrfs:
Not sure how to proceed here in terms of branching, but Im currently doing as you requested last time @jwinarske, basing it on `scarthgap`. But let me know if you need it to be `master` instead.

jwinarske:
merging anyway, as it just makes common sense ;)  Thanks!

leojrfs:
> @leojrfs Where you seeing runtime problems without this change?
Yes, a lot of random problems. A short description of the issues and the reason explained in the description of this PR.

> What does `extra_ldflags` expand to at `do_compile`?
Now, after this PR, just to the default yocto-provided target LDFLAGS. As it should be.

PR #692, opened 2025-08-19
Fix builds for multiple runtime modes [YOCIMX-9102]

thochstein:
A build for multiple runtime modes doesn't work properly because each mode uses the same build folder as the others and thus overwrites the output of the previous mode build.


jwinarske:
@thochstein 

The building of apps was intentionally left out for everything other than runtime=release (AOT).  The development workflow is host driven; you drive the debug work on a target from a host machine.

Running a standalone debug image has little benefit, other than it runs much slower than an AOT image.

Profiling runtime is not currently supported by flutter custom-devices; it has to be hand run.  One creates an AOT on the dev using workspace-automation command.

This PR would increase my time to smoke test releases, without a clear benefit.

PR #696, opened 2025-08-28
fix building workspace packages

ardera:
Hey @jwinarske ,

dart/flutter recently introduced workspaces which AFAIU, basically makes dart treat a mono-repo containing a lot of packages, like it was a single package. That speeds up analysis a bit.

It does change some things in the build, more precisely the `package_config.json` is now not in the directory of the app that should be built, but in the workspace root.

Luckily dart creates a `workspace_ref.json` file that indicates where the `package_config.json` can be found.

This patch makes workspace packages build fine for us, without any other changes in the app recipes.

Can also create PRs for the other yocto releases.

### DIFF
--- a/conf/include/common.inc
+++ b/conf/include/common.inc
@@ -403,6 +403,7 @@ def versiontuple(v):
 def build_app(d, source_dir, flutter_sdk, pubspec_appname, flutter_application_path, staging_dir_target, source_root, env):
     import glob
     import shutil
+    import json
 
     # determine build type based on what flutter-engine installed.
     datadir = d.getVar('datadir')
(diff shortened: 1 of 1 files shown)


jwinarske (approved):

ardera:
Okay so this is an example of a workspace app that fails to build without the fix:

app:
[workspace-app.zip](https://github.com/user-attachments/files/22079430/workspace-app.zip)

recipe:
```bitbake
LICENSE = "CLOSED"
LIC_FILES_CHKSUM = ""

SRC_URI = "file:///.../workspace-app.zip"

inherit flutter-app

PUBSPEC_APPNAME = "myapp"
FLUTTER_APPLICATION_PATH = "myapp"
```

PR #698, opened 2025-09-16
flutter-engine_git: Add ccache support

markyang92:
Adds ccache support to the GN build when inheriting from the ccache class.


jwinarske:
Hi @markyang92,

what kind of performance gain do you see running ccache building flutter-engine?

markyang92:
Hi @jwinarske 

My build environment is as follows
System: Ubuntu 22.04 x86_64
CPU: AMD Ryzen 9 7950x 16-core 32-Thread
RAM: 128GB
Local Storage: SSD

### Test 1: Without ccache
```bash
# clone poky
$ source oe-init-build-env
$ echo "MACHINE = \"genericarm64\"" >> build/conf/local.conf

# clone openembedded-core
# clone meta-flutter (HEAD is my commit c5724da8f4b63ccf60baec700a2556240a4030e0)
# Added meta-oe and meta-flutter to BBLAYERS in build/conf/bblayers.conf
$ bitbake flutter-engine

$ cat build/tmp/buildstats/20250917051805/flutter-engine-3.32.8-r0/do_compile | grep "Elapsed time"
Elapsed time: 1433.78 seconds
```


### Test 2: With ccache first time (No ccache data available)
```bash
# set ccache settings
$ echo "INHERIT += \"ccache\"" >> build/conf/local.conf
# By default, other components disable ccache.
$ echo "CCACHE_DISABLE = \"1\"" >> build/conf/local.conf
# Enable ccache only for flutter-engine.
$ echo "CCACHE_DISABLE:pn-flutter-engine = \"0\"" >> build/conf/local.conf
# For ccache stat debug
$ echo "export CCACHE_STATSLOG = \"${T}/ccache_stats.log\"" >> build/conf/local.conf

$ bitbake flutter-engine -c clean && bitbake flutter-engine -C do_compile
cat build/tmp/buildstats/20250917061911/flutter-engine-3.32.8-r0/do_compile | grep "Elapsed time" 
Elapsed time: 872.70 seconds
```
- Even on the first build using ccache, it is faster than without ccache.
- Let's look at the logs.
```bash
$ cd build/tmp/work/armv8a-poky-linux/flutter-engine/3.32.8/temp 
$ export CCACHE_STATSLOG="./ccache_stats.log"
$ ../recipe-sysroot-native/usr/bin/ccache --show-log-stats
Cacheable calls:   28814 / 28853 (99.86%)
  Hits:            17736 / 28814 (61.55%)
    Direct:        15208 / 17736 (85.75%)
    Preprocessed:   2528 / 17736 (14.25%)
```
- Because it is both generating the ccache and using ccache data at the same time.


### Test 3: With ccache second time (ccache data available)
```bash
$ bitbake flutter-engine -c clean && flutter-engine -C do_compile
$ cat build/tmp/buildstats/20250917081102/flutter-engine-3.32.8-r0/do_compile | grep "Elapsed time"
Elapsed time: 299.87 seconds

$ cd build/tmp/work/armv8a-poky-linux/flutter-engine/3.32.8/temp 
$ export CCACHE_STATSLOG="./ccache_stats.log"
$ ../recipe-sysroot-native/usr/bin/ccache --show-log-stats
Cacheable calls:   28814 / 28853 (99.86%)                                                                                                                                                                                                                                                                            
  Hits:            28814 / 28814 (100.0%)                                                                                                                                                                                                                                                                            
    Direct:        28760 / 28814 (99.81%)
    Preprocessed:     54 / 28814 ( 0.19%)
```

-|do_compile time (s)
:---|:---
Without ccache | 1433.78
With ccache first time (No ccache data)| 872.70
With ccache second time (ccache data available)| 299.87

jwinarske:
@markyang92 that looks great.  What about building actual image, say core-image-Weston or core-image-minimal; same stats with/without ccache?  Curious how it saturates the machine.

jwinarske:
Seems I would need to add a new CI job altogether in order to test regressions.

Might be a good time to add mold linker support.

Same for Musl; no tests.

jwinarske:
@markyang92 in your tests be sure to account for engine archive fetching.  A -cdo_cleanall is what deletes the archive.  A subsequent run includes the archive fetching.  All runs after use cached archive.

markyang92:
ccache is a way to speed up compilation during frequent do_compile tasks when ccache data is available. 
If there are no changes, ccache is not used because of sstate-cache hits (the do_compile itself is skipped). 

So I forcibly removed the workdir with -c clean and forced do_compile with -C do_compile, and measured only the compile time.

Q.  What about building actual image, say core-image-Weston or core-image-minimal; same stats with/without ccache?
A. When building the image, by default if there are no changes in flutter-engine the do_compile task is skipped (sstate cache hit). If there are changes to flutter-engine and run bitbake <image>, do_compile will of course be executed using ccache. 
In that case, the do_compile for flutter-engine will naturally be faster thanks to ccache.
The compile speed of flutter-engine is naturally similar to the above results.

Q.  A -cdo_cleanall
Using do_cleanall is somewhat problematic. 
If you look at `poky/meta/classes:62`, when `cleanall` is used it removes the ccache data, so any measurements taken after using `do_cleanall` are always done with no ccache data present.

recipes-graphics/flutter-engine/files/0001-BUILD.gn-Support-LTO-build-on-Linux.patch
```diff
@@ -0,0 +1,26 @@
+From e9917111c1d8bb0ca1c9319aaf86b56880bbddee Mon Sep 17 00:00:00 2001
```
jwinarske, line 1: remove from this PR

recipes-graphics/flutter-engine/flutter-engine_git.bb
```diff
@@ -46,6 +46,7 @@ SRC_URI = "\
     file://0001-memcpy-void-cast.patch \
     file://0001-swiftshader-pointer-cast-to-void.patch \
     file://BUILD.gn.in \
+    file://0001-BUILD.gn-Support-LTO-build-on-Linux.patch \
```
jwinarske, line 49: remove from this PR

jwinarske (changes requested):

PR #699, opened 2025-09-29
flutter-engine_git: Support LTO build on Linux

markyang92:
Enable LTO for the is_linux case in BUILD.gn as well.

- Although `--lto` is configured in `flutter-engine_git.bb`, `-flto` is not actually applied during the build.

```
# meta-flutter-recipes-graphics/flutter-engine/flutter-engine_git.bb
139: GN_ARGS = "\
140:     ${PACKAGECONFIG_CONFARGS} \
141:     --clang --lto \

but

# Check whether LTO has been applied in compile_commands.json
$ cat build/tmp/work/armv8a-poky-linux/flutter-engine/3.32.8/sources/gn/engine/src/out/linux_debug_arm64/compile_commands.json | grep "\-flto"

..nothing
```

- After applying this patch, I can find `-flto` in `compile_commands.json` when building.

```
$ bitbake flutter-engine

# Check whether LTO has been applied in compile_commands.json
$ cat build/tmp/work/armv8a-poky-linux/flutter-engine/3.32.8/sources/gn/engine/src/out/linux_debug_arm64/compile_commands.json | grep "\-flto"
    "command": ".....clang++ ... Omitted .. \
 -flto ..... \
-g0 -Wunreachable-code  -fvisibility-inlines-hidden -std=c++17 -fno-rtti -nostdinc++ -nostdinc++ -fvisibility=hidden -fno-exceptions   -c ../../flutter/fml/string_conversion.cc -o  clang_x64/obj/flutter/fml/string_conversion.string_conversion.o"
```

### DIFF
--- a/recipes-graphics/flutter-engine/flutter-engine_git.bb
+++ b/recipes-graphics/flutter-engine/flutter-engine_git.bb
@@ -138,7 +138,7 @@ CLANG_PATH:riscv64 = "${STAGING_DIR_NATIVE}/usr"
 
 GN_ARGS = "\
     ${PACKAGECONFIG_CONFARGS} \
-    --clang --lto \
+    --clang \
     --no-goma --no-rbe \
     --no-enable-unittests \
     --no-stripped \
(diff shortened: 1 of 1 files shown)


markyang92:
I'll check. The following results regarding the use of LTO were reported by Flutter upstream.
from: https://github.com/flutter/flutter/issues/177660


LTO produced a significantly smaller binary for `libflutter_engine.so`.  But the `flutter_tester` binary was larger with LTO.

binary | with LTO | without LTO
:-- | :-- | :--
libflutter_engine.so | 33.4 MB | 42.7 MB
flutter_tester | 34.5 MB | 33.2 MB

jwinarske:
@markyang92 yes that's expected, no surprise there.  I think we want it as a Yocto package config option, so it can be turned off depending on the circumstance; specific LLVM toolchain version that segfaults with LTO enabled.

I think the only critical use case would be for runtime=release.

jwinarske:
closing in favor of https://github.com/meta-flutter/meta-flutter/pull/713

PR #739, opened 2025-12-12
gn.py: use protocol not proto in SRC_URI

shr-project:
It uses https by default, but if someone uses e.g. ssh, then using "protocol" in SRC_URI doesn't work with gn:// but "proto" is deprecated since bitbake 1.16 from 2012 and each recipe using this triggers multiple warnings from:
https://git.openembedded.org/bitbake/commit/?h=1.16&id=53e6b630f0463d2d07cdaa9c9eb36794dc9b6b69


jwinarske:
@shr-project sometimes the maintainer mule needs a mental break.  This one has been a totally thankless job.

PR #740, opened 2026-02-03
switch to fork of Sony embedded Linux embedder

lhoward:
Sony is no longer maintaining their embedded Linux embedder (sony/flutter-elinux#289). Switch to a community fork, which supports Flutter 3.38.3.

cc @martinetd

### DIFF
--- a/recipes-graphics/sony/sony-flutter.inc
+++ b/recipes-graphics/sony/sony-flutter.inc
@@ -4,8 +4,8 @@
 
 SUMMARY = "Embedded Linux embedding for Flutter"
 AUTHOR = "Sony Group Corporation"
-HOMEPAGE = "https://github.com/sony/flutter-embedded-linux"
-BUGTRACKER = "https://github.com/sony/flutter-embedded-linux/issues"
+HOMEPAGE = "https://github.com/flutter-elinux/flutter-embedded-linux"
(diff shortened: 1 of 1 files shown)


martinetd:
We didn't discuss this part much yet :sweat_smile: 

So far there's just @lhoward and me as admins of the group, and as far as I can tell we're both just some company using flutter-elinux (I'm from https://github.com/atmark-techno and he's https://github.com/PADL -- I can't speak for PADL but we're maintaining a software stack for our embedded boards which happened to use flutter elinux, so we're more like "enablers" than direct users and we already have plenty of customers using it so we don't want to switch and will likely stick around for a while. All of our work is open source so there's practically no risk of going dark all of a sudden either)

I think we'll be mostly done with the backlog of PRs around next week, and I planned tackling organization a bit more squarely after that point, but honestly I have no great plan as long as the project won't die if I get hit by a bus, and so far @lhoward has been very helpful so I'm confident things will be mostly ok as long as either of us is around.. But I definitely agree it'd be better to formalize.
(Sony also gave the OK to fork and said they'd point to the new repo in their readme, so I planned to work on that at that point)

FredeHoey:
I don't know how I missed this when I created #741. The patch is essentially the same, but I've renamed the parent directory, to match the new org.

jwinarske:
Closing this in favor of https://github.com/meta-flutter/meta-flutter/pull/758, as it captures the deprecation warning I was looking for.

PR #741, opened 2026-02-05
Sony's embedder moved to the flutter-elinux org

FredeHoey:
See:
1. https://github.com/sony/flutter-embedded-linux/pull/443#issuecomment-3815213824
2. https://github.com/flutter-elinux/flutter-embedded-linux


FredeHoey:
@jwinarske I just added the parallel recipe idea you proposed in #740. It's not my PR though, so I suggest we merge this one #758

PR #744, opened 2026-03-12
common.inc: Add logic to enable/disable pub when building app

Antonios-C:
Dynamic logic checking if network is enabled for building app, enabling/disabling pub for flutter build.

### DIFF
--- a/conf/include/common.inc
+++ b/conf/include/common.inc
@@ -418,6 +418,12 @@ def build_app(d, source_dir, flutter_sdk, pubspec_appname, flutter_application_p
         bb.note(f'Using new build scheme')
         new_build_scheme = True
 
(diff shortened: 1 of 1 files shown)


Antonios-C:
@jwinarske  yeah I agree ! 

My suggestion: 

- Add  ``` --no-pub``` option to be default included within ```flutter_build_args```
- Add a variable (example: ```FLUTTER_NETWORK_BUILD```) that when enabled warns users of their actions 
   -  In the background, within a class it would remove the ```--no-pub``` option from ```flutter_build_args``` and enable ```do_compile[network]```

Let me know your thoughts ?

jwinarske:
@Antonios-C this works for me as-is.  It tracks the logical state correctly.  Independent of any warnings.

PR #755, opened 2026-04-24
flutter-apps: wonderous-app-wonders: Fix source directory for git fetcher

BoergeSt:
The git fetcher unpacks sources into ${WORKDIR}/git, but S was not set accordingly. This causes do_compile to fail due to an invalid source directory and prevents LICENSE from being found during license checks.

Set S = "${WORKDIR}/git" to fix the source path.


jwinarske:
@BoergeSt This does not apply to master.

PR #756, opened 2026-05-04
[scarthgap] flutter-engine: Use libdir when locating target libraries

dpmin7:
The clang-layer bbappend hardcodes the target library path as ${STAGING_DIR_TARGET}/usr/lib when creating the crtbeginS.o symlink.

This does not work in multilib builds where the target library directory can be different, such as /usr/lib32 or /usr/lib64. Use ${libdir} instead so the path follows the selected target ABI.

This fixes the lib32 flutter-engine build failure caused by looking for crtbeginS.o in the wrong library directory.


dpmin7:
@jwinarske <
This has already been resolved on the master branch. I believe this minor change should be applied to all currently active branches. Could you please review it when you have some time?
Thank you.

jwinarske:
@dpmin7 Good find, I like this solution.  This would need to go into `dunfell`, `kirkstone`, and `scarthgap`.

jwinarske (approved):

PR #758, opened 2026-05-11
Sony embedder moved to new org

FredeHoey:
#740 seems stale and #741 is closed, so I'm opening this new PR to update the location of the sony embedder including warnings about deprecation of the sony recipes.

The FLUTTER_ELINUX_USE_SONY variable controls wheter or not the build uses sony's or flutter elinux's recipes. 

- When FLUTTER_ELINUX_USE_SONY == 1, show 

```
WARNING: sony flutter-embedded-linux recipes are deprecated. Use recipes-graphics/flutter-elinux/ instead. FLUTTER_ELINUX_USE_SONY=1 currently set - remove to silence this warning.
```
- When FLUTTER_ELINUX_USE_SONY is unset, use sony recipes and show
```
WARNING: FLUTTER_ELINUX_USE_SONY is unset. Defaulting to flutter-elinux recipes. Set FLUTTER_ELINUX_USE_SONY = "0" to silence this warning, or FLUTTER_ELINUX_USE_SONY = "1" to use the deprecated sony recipes.
```
- When FLUTTER_ELINUX_USE_SONY == 0, Do nothing, use flutter-elinux recipes


jwinarske:
@FredeHoey looks good.  How did you test this?

lhoward:
Just checking the DRM EGLStream backend needs the Wayland dependency? (I've not used it so it might, just seems odd on first inspection.)

jwinarske:
@lhoward no idea.  I know it’s not needed with the new IVI-homescreen drm/kms-EGL backend.

FredeHoey:
> @FredeHoey looks good. How did you test this?

I built all the recipes provided by both repositories with

FLUTTER_ELINUX_USE_SONY = "1"

FLUTTER_ELINUX_USE_SONY = "0"

And with it unset, and it passed. I only use the drm gbm backend, so I haven't tested all programs at runtime, but if they worked  from Sonys tree, I don't see what would have broken them in the fork.

> Just checking the DRM EGLStream backend needs the Wayland dependency? (I've not used it so it might, just seems odd on first inspection.)

I'll double-check when I'm back at my desk :)

FredeHoey:
>> Just checking the DRM EGLStream backend needs the Wayland dependency? (I've not used it so it might, just seems odd on first inspection.)
>
> I'll double-check when I'm back at my desk :)

It appears that it doesn't. I've also updated SRCREV to latest master.

@jwinarske, I just want to note that a lot of the sony recipes don't compile under GCC 15 which is the default gcc compiler for wrynose, so if the flag is not set, the warning will possibly drown in compiler output.

jwinarske:
@FredeHoey if the don’t build in wrynose, they won’t build in master.  Any recipe that doesn’t build will be removed.  I started roll to 3.44.2 yesterday.  I’ll take a look at the Sony recipes next week.  Also want to compare them to the new high performance backends in IVI-homescreen.  If they are dismal in comparison, likely I will drop them.

FredeHoey:
@jwinarske I can kick off a build of all the original Sony recipes to see which ones don't build on master/wrynose and remove them if you want?

EDIT: build came back almost all red, only the video-player-plugin from the sony recipes builds

- [ ] flutter-drm-eglstream-backend
- [ ] flutter-drm-gbm-backend
- [ ] flutter-external-texture-plugin
- [X] flutter-video-player-plugin
- [ ] flutter-wayland-client
- [ ] flutter-wayland-so
- [ ] flutter-x11-client

All recipes from flutter-elinux compile green

FredeHoey:
@jwinarske I just opened another merge request based off of this one, that removes the sony recipes. We can merge this one or both. Whatever you prefer :)

FredeHoey:
@jwinarske Do you have any further comments for this before it can be merged? 🙏🏼

jwinarske:
@FredeHoey Thanks for picking this up after #740/#741, and for the build matrix you posted — that settled the question of whether the Sony tree is worth keeping.

I checked out the branch and reviewed against the merge-base (`f816602`). Notes:

**Mechanism is sound.** The `BBMASK` toggle in `layer.conf` works even though layer.conf parses before local.conf — `+=` with a `${@...}` expression is stored unexpanded and BBMASK is evaluated after the full config parse, so `FLUTTER_ELINUX_USE_SONY` from local.conf is visible by then. The tri-state warning behavior (unset → nudge, `"1"` → deprecation warning, `"0"` → silent) is coherent, and unlike in #765, the `sony-flutter.inc` warning text is accurate here since the sony tree only parses when the flag is actually set to 1.

**Verified:** `SRCREV a98514bf` is reachable in `flutter-elinux/flutter-embedded-linux` and the LICENSE md5 matches at that rev. Dropping `0001-Add-missing-stdint-header.patch` from the new external-texture-plugin copy is correct — it's addressed at the new rev.

**One bug — the EGLStream wayland fix landed on the wrong copy.** Per @lhoward's question you removed `REQUIRED_DISTRO_FEATURES += "wayland"` from `recipes-graphics/sony/flutter-drm-eglstream-backend_git.bb`, but the new `recipes-graphics/flutter-elinux/` copy — the one that's active by default — still has it:

```bitbake
# recipes-graphics/flutter-elinux/flutter-drm-eglstream-backend_git.bb
REQUIRED_DISTRO_FEATURES += "wayland"   # <-- should be dropped here too
```

Since you confirmed the DRM-EGLStream backend doesn't actually need wayland, please remove it from the flutter-elinux copy (the sony copy barely matters at this point). This also affects #765, where the rename carries the wayland requirement forward and the sony-side fix is discarded.

**On #758 vs #765:** the commits here (`710b9f9`, `a8f2368`) are literally the first two commits of #765, so #765 strictly supersedes this PR. Given your build results — 6 of 7 Sony recipes fail under GCC 15 on wrynose — a deprecation cycle would mostly be preserving a toggle to code that doesn't compile. And as you noted, the deprecation warning would drown in compiler errors for anyone who actually flips the flag. My preference is to skip the transition period: close this one and land #765 directly, with two fixups there —

1. Carry the EGLStream wayland removal into the flutter-elinux recipe.
2. Resolve the orphaned `sony-flutter.inc` (see my comments on #765 — delete it or turn it into a warn-and-forward shim; as-is the warning references a flag nothing reads anymore).

If you'd rather keep a one-release transition window, this PR is mergeable once the EGLStream parity fix is in — but I don't think the Sony tree earns it.

PR #762, opened 2026-06-11
appstream-dart-flathub-catalog: skip native build via hooks user-define

jwinarske:
## Summary

The `SKIP_NATIVE_BUILD` env var never reached appstream_dart's native-asset hook, since Dart runs hooks in a hermetic environment. This switches to the `skip_native_build` hooks user-define, injected into `pubspec.yaml` by `do_inject_user_defines` (the same channel already used for sqlite3), so the hook stands down and `do_cmake_compile` remains the single producer of `libappstream.so` for the target.

Requires appstream_dart >= `ff1840c` (PR #10), which reads the user-define; SRCREV is bumped accordingly.

Also included: revert of the earlier libsqlite3-0 dependency/symlink change.

## Test plan

- Build the recipe and confirm `libappstream.so` is produced solely by `do_cmake_compile` (cross toolchain), with the native-asset hook standing down.


jwinarske:
Wrong target branch — this should not go to master.

PR #765, opened 2026-06-13
Remove sony flutter recipes

FredeHoey:
Follow-up to #758. 

Remove the sony recipes for flutter-elinux, as the project is unmaintained and the recipes don't build


FredeHoey:
@jwinarske any comments on this one?

FredeHoey:
Hi @jwinarske, If you have time to look at this or just #758, I would very much appreciate it 🙏🏼

jwinarske:
@FredeHoey Thanks for the follow-up to #758, and sorry for the delay.

I went through all three commits. The core of this is good:

- Recipe file names (PNs) are unchanged — only the directory moved from `recipes-graphics/sony/` to `recipes-graphics/flutter-elinux/`, so existing `.bbappend`s and `IMAGE_INSTALL` entries keep working.
- Verified `SRCREV a98514bf` is reachable in `flutter-elinux/flutter-embedded-linux`, and the `LIC_FILES_CHKSUM` md5 (`d45359c8...`) matches the LICENSE at that rev — no checksum churn.
- Dropping `0001-Add-missing-stdint-header.patch` along with the version bump makes sense given the gcc fixes in the new rev.
- `HOMEPAGE`/`BUGTRACKER` correctly updated to the flutter-elinux org.
- Grepped the tree at the PR head — no dangling references to `recipes-graphics/sony/` anywhere else (workflows, docs, kas).

Two things before merge:

**1. `sony-flutter.inc` is now orphaned, and the deprecation warning is misleading**

After the move, nothing in-tree `require`s `recipes-graphics/sony/sony-flutter.inc`, and `FLUTTER_ELINUX_USE_SONY` isn't read anywhere in the layer. So:

- The `bb.warnonce` in the `__anonymous` function only fires if someone `require`s the file directly, and
- The message *"FLUTTER_ELINUX_USE_SONY=1 currently set - remove to silence this warning"* will print regardless of whether that variable is set, since nothing gates on it.

Given the PR title is "Remove sony flutter recipes", I'd just delete `recipes-graphics/sony/` entirely. If you want a transition shim for downstream layers that require the .inc by path, reduce it to:

```bitbake
python __anonymous() {
    bb.warnonce("recipes-graphics/sony/sony-flutter.inc is deprecated; "
                "use recipes-graphics/flutter-elinux/flutter-elinux.inc instead.")
}

require recipes-graphics/flutter-elinux/flutter-elinux.inc
```

That way downstream builds keep working against the new org/SRCREV instead of the stale Sony pin, and the warning text is accurate. My preference is outright deletion — clean break, and the rename commits preserve history.

**2. Commit message nit**

The final commit says the recipes are removed, but as it stands `sony-flutter.inc` is retained/deprecated. Once item 1 is resolved (either direction), please make the commit message match the actual behavior.

**Optional nit:** `AUTHOR = "Sony Group Corporation"` in `flutter-elinux.inc` — `AUTHOR` has been dropped from oe-core; fine to remove while you're in there, but not blocking.

Happy to merge once the `sony-flutter.inc` question is settled. Also confirming the direction here: Sony's original repo is unmaintained and the flutter-elinux org fork is the live continuation, so pointing the layer at it is the right call.

FredeHoey:
@jwinarske Thanks for the feedback!

I've rebased the patch so the first commit is simply the move and the second is a version upgrade to latest. I also removed (which is now just considered a rename by git) sony.inc as well as the AUTHOR field

PR #768, opened 2026-06-15
sony-flutter: replace EXCLUDE_FROM_SHLIBS with PRIVATE_LIBS

markyang92:
Backport of #764 to scarthgap.

The EXCLUDE_FROM_SHLIBS / do_package_qa lines are present and the engine
installs per mode, so dropping EXCLUDE_FROM_SHLIBS naively trips
"Multiple shlib providers for libflutter_engine.so". PRIVATE_LIBS
skips that soname and restores the rest of the scan.

### DIFF
--- a/recipes-graphics/sony/sony-flutter.inc
+++ b/recipes-graphics/sony/sony-flutter.inc
@@ -69,7 +69,6 @@ do_configure:prepend() {
     ln -sf ${FLUTTER_ENGINE_PATH}/lib/libflutter_engine.so ${S}/build/libflutter_engine.so
 }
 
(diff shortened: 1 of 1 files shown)


jwinarske:
@markyang92 have you tested this on scarthgap?  Does this affect kirkstone and dunfell as well?

markyang92:
I reproduced the same behavior on a `poky scarthgap` based setup.
I will also try `kirkstone`.
`dunfell` has been EOL for a long time, so I'm not sure if it is still needed.

jwinarske (approved):

PR #773, opened 2026-08-20
roll to Flutter 3.47.1 and add ivi-homescreen v3

jwinarske:
Rolls the SDK to Flutter 3.47.1 (Dart 3.13.1) and adds ivi-homescreen v3
support, along with the recipes and class needed to build against it.

## SDK roll

`tools/roll_meta_flutter.py` to 3.47.1. Packages depending on `hooks ^2.1.0`
need `meta >= 1.19.0`, which cannot resolve against the `meta 1.17.0` that
3.38.3 pins, so the bump is a prerequisite for the app recipes below.

## ivi-homescreen v3

`ivi-homescreen_3.0.bb` and `flutter-auto_3.0.bb` over a shared
`ivi-homescreen-v3.inc`. Backends are no longer mutually exclusive in v3 --
wayland-egl, wayland-vulkan, wayland-leased-drm, drm-kms-egl, drm-kms-vulkan,
software, headless-egl and headless-vulkan may all be compiled in, with the
active one resolved at runtime.

- Wayland dependencies moved onto the Wayland backends, so a DRM/KMS, software
  or headless build no longer requires the wayland distro feature. The
  leased-DRM backend omits `wayland-protocols`: it has no `wl_surface` and reads
  nothing from the package.
- The plugins tree and the v4l2-webrtc-codec encoder sources are fetched only
  when built, so `disable-plugins` means no plugins clone and no sd-bus
  dependency.
- A parse-time consistency matrix rejects the combinations upstream CMake
  rejects, and warns on options that would otherwise silently do nothing.
- `plugins/webrtc` is deliberately not exposed: it links libwebrtc's C++ API
  into the embedder and its vendored flutter-webrtc copy has drifted.

`ivi-homescreen-shared` builds the `shared/` subtree standalone. Both embedders
install an identical `libihs_shared`, so a single owner lets them be installed
side by side, and consumers can depend on the C-ABI library without pulling in
an embedder.

`wayland-cxx-scanner` provides the native code generator v3 needs to generate
its Wayland bindings on the build host.

## Flutter apps with native assets

`flutter-app-native.bbclass` covers apps whose Dart packages carry a
native-assets build hook driving CMake. The hook runs inside `flutter build`
and cannot see the OE cross environment, so the class puts a cmake wrapper on
PATH that injects the toolchain file and pkg-config settings, then installs the
resulting libraries into the app bundle. `FLUTTER_NATIVE_VERBOSE` surfaces the
hook output, which the native-assets builder otherwise swallows.

New: bluez_native flutter_ble_scanner, flatpak_dart flutter_remote_manager,
packagekit_dart packagekit_catalog. The appstream_dart flathub_catalog example
is repointed at flatpak-minimal and moved onto the class, replacing its
hand-rolled cmake and flutter-app hybrid.

## libwebrtc

Restored and updated: absorbs the v4l2-webrtc-codec hardware decoder as a gn
source_set, drives the audio and X11 backends from DISTRO_FEATURES (when both
wayland and x11 are present the pipewire path wins and the libX* dependencies
drop out), and stamps a versioned `DT_SONAME` so consumers record
`libwebrtc.so.144`. Desktop capture is optional but requires pipewire: it adds
virtual methods to the public interfaces, so a consumer built without
`RTC_DESKTOP_DEVICE` would get a mismatched vtable.

## flutter-engine

- Drop patches upstream absorbed or superseded, including the two libc++ musl
  patches now covered by `-D_LIBCPP_HAS_MUSL_LIBC`.
- Several long-standing seds had silently stopped matching after upstream
  reworded the files they edit. Pass `--no-default-linux-sysroot` rather than
  editing `sysroot.gni`, and match the `vulkan_headers.gni` assignment however
  it is spelled.
- Set `dart_include_wasm_opt=false`, which upstream applies only to host builds
  and without which the wasm-opt link fails on duplicate libc++abi symbols.
- Split the runtime modes into their own packages so an image carries only the
  mode it runs. `flutter-engine` remains a metapackage over the modes built.
- Add an ABI interop gate over `libflutter_engine.so`: no dynamically linked C++
  runtime, a libc matching the target, a static unwinder, no exported or
  undefined C++ symbols. The engine is built with clang and an in-tree libc++
  but is dropped into GCC userlands, where the only shared contract is the C
  embedder ABI over the target libc.

## Testing

Built on x86-64 (glibc, wayland + x11 + pipewire + polkit):

- `ivi-homescreen`, `flutter-auto`, `ivi-homescreen-shared` -- clean; the
  embedders carry no overlapping files and both resolve `ivi-homescreen-shared`.
- All four app recipes -- clean.
- `libwebrtc` -- clean across the option matrix; all 32 PACKAGECONFIG
  combinations verified at parse level (24 valid, 8 correctly rejected), with
  the all-on and all-off extremes built.
- `core-image-weston` with ivi-homescreen and the four apps installed --
  clean, all five present in the rootfs manifest.

Not covered: musl (reasoned from emb_cli's working implementation, not built),
and nothing has been booted.


lhoward:
Will it be possible to restore `wrynose` compatibility?

lhoward:
I needed this patch to get past fetch.

```diff
diff --git i/recipes-graphics/flutter-engine/flutter-engine_git.bb w/recipes-graphics/flutter-engine/flutter-engine_git.bb
index f5fd9c17..3941eabe 100644
--- i/recipes-graphics/flutter-engine/flutter-engine_git.bb
+++ w/recipes-graphics/flutter-engine/flutter-engine_git.bb
@@ -68,7 +68,12 @@ SRC_URI:libc-musl += "\
 
 inherit gn-fetcher features_check pkgconfig
 
-GN_DEPS_SED_PATCHES:pn-flutter-engine:aarch64 = "# Always download the JDK since java is required for running the formatter.|'condition': 'download_android_deps',"
+GN_DEPS_SED_PATCHES:pn-flutter-engine:aarch64 = "not (host_os|download_android_deps and not (host_os"
```

lhoward:
Rebased onto `f458ad25` — thanks for that, and for catching `meta-flutter-apps/conf/layer.conf` too, which I had missed.

We now carry the fix below in our own bbappend rather than patching this layer, so we build against your branch unmodified. Posting it as a suggestion rather than a request.

```
--- a/recipes-graphics/flutter-engine/flutter-engine_git.bb
+++ b/recipes-graphics/flutter-engine/flutter-engine_git.bb
@@ -68,7 +68,7 @@ SRC_URI:libc-musl += "\
 
 inherit gn-fetcher features_check pkgconfig
 
-GN_DEPS_SED_PATCHES:pn-flutter-engine:aarch64 = "# Always download the JDK since java is required for running the formatter.|'condition': 'download_android_deps',"
+GN_DEPS_SED_PATCHES:pn-flutter-engine:aarch64 = "not (host_os|download_android_deps and not (host_os"
```

---

*Claude-generated (Opus 5), from building this branch rather than reading the diff.*

**1. `GN_DEPS_SED_PATCHES` produces invalid DEPS on 3.47.1.** The openjdk cipd entry gained its own condition upstream *and kept* the comment the rewrite targets, so the injected key collides and `do_fetch` dies in `gclient_eval`:

```
ValueError: duplicate key in dictionary: condition (file '.../sources/gn/./DEPS', line 630)
```

```python
'engine/src/flutter/third_party/java/openjdk': {
   'condition': 'not (host_os == "linux" and host_cpu == "arm64")',   # upstream, new in 3.47.1
   'condition': 'download_android_deps',                              # injected by the rewrite
```

This is a hard failure on any aarch64 target, not a missed optimisation — the rewrite corrupts `DEPS` before gclient evaluates any condition. Confirmed by parking the source tarball and forcing `do_fetch` against a pristine tree.

Folding `download_android_deps` into the upstream expression keeps the JDK unfetched without a second key. Deleting `GN_DEPS_SED_PATCHES` outright would work too, since 3.47.1's `DEPS` now carries the arm64-host guard the rewrite existed for — the cost is that x86-64 hosts fetch a JDK nothing opens, and it lands in the cached source tarball. Either resolves it; leaving it as-is does not.

The anchors are deliberately quote-free: `GN_DEPS_SED_PATCHES` is a plain bitbake `"…"` variable and bitbake does not unescape `\"`.

**2. `GN_DEPS_SED_PATCHES` is in no task signature**, which is worth fixing regardless of how you resolve the above:

```
do_fetch[vardeps] += "GN_DEPS_SED_PATCHES"
```

It is read by `lib/gn.py` at `d.getVar`, never from a recipe task, so bitbake's dependency scan never sees it. Editing it invalidates nothing. My first attempt to test a change to it ran **0 tasks** and reported `Tasks Summary: ... all succeeded` — the fetcher had short-circuited on the cached tarball and nothing signalled that the edit was inert. Anyone tuning that variable gets a stale result and a green build.

**3. A retained partial sync can never be resumed.** `clean()` deliberately keeps `ud.syncpath` when a gclient sync fails:

> *If an error occurs during the gclient sync, we aim to retain the partial sync directory ... The sync directory can grow up to 14GB.*

But `basecmd` opens with `git init . && git remote add origin …`, which fails on any populated tree:

```
error: remote origin already exists.
ERROR: Bitbake Fetcher Error: FetchError('Unable to fetch URL from any source.', 'gn://github.com/flutter/flutter.git')
```

So the 14GB is preserved and then unusable — the retry fails before it reaches gclient. Making the remote setup idempotent would make the retention do what the comment intends:

```
git remote add origin <uri> || git remote set-url origin <uri>
```

This is more visible after `gn-fetcher: key the fetch on the gclient config`: `do_fetch` now re-runs on a config change, and on any existing build tree it lands on the tree `do_unpack` left behind and fails there. I hit it that way on 3.47.1, and separately after a failed fetch — same cause both times.

---

**Follow-up on 3: the resumable-sync fix needs a companion change, and the suggestion I gave was incomplete.**

With `1a53888a` in, `do_fetch` no longer fails on a populated `syncpath` — but `S` is also the build directory, so after a successful build that directory holds `engine/src/out/<config>/` as well as sources. The fetch now succeeds against a *built* tree, and `packcmd` tars it wholesale:

```python
ud.packcmd = f'tar -I "pbzip2 ..." -cf {ud.localpath} ./; ...'
```

Observed here after `547b4936` changed the cache key and forced a refetch:

```
flutter-engine-3.47.1-<srcrev>.tar.bz2               2.2 GB
flutter-engine-3.47.1-<srcrev>-a9b3709eda82.tar.bz2  5.2 GB   <- 4.8 GB of engine/src/out
```

The previous behaviour was a loud failure; this is silent, and it caches build output as source. That seems worth closing before `jw/gn-premirror` lands, since a poisoned tarball could reach a shared mirror.

Excluding the build output from the pack would fix it independently of how the tree got dirty:

```
tar ... --exclude=./engine/src/out -cf {ud.localpath} ./
```

There is also a latent correctness edge: an unpacked tarball containing `out/` gives ninja prebuilt targets, so `do_compile` could no-op and ship artifacts from an earlier build. It did not happen here — ninja ran all 13927 steps — but it is reachable.

lhoward:
Correction on the rule 4 report that likely prompted `be7b7b72` — it was our fault, not glibc's.

We were overriding `CLANG_PATH = "${STAGING_DIR_NATIVE}/usr"` in a local bbappend, unconditionally rather than only for riscv64. That put an aarch64 target on the native clang, which is exactly the `--rtlib=libgcc` case your commit describes. Dropping the override puts us back on the bundled clang and the gate passes unexempted:

```
                 override in place                     override removed
NEEDED   libfontconfig, libm, libgcc_s, libc, ld   libfontconfig, libpthread, libm, libc, ld
rule 4   native_clang note (exempted)              NOTE: ABI gate passed for 1 library(ies)
```

Full image build, clean both ways. So `be7b7b72` is right, but the exemption looks riscv64-only in practice — a stock glibc/aarch64 consumer does not trip rule 4, and my earlier comment implying otherwise was wrong.

Same override was also why `-fuse-ld=lld` could not find an `lld` for us and why a target link carried `-L${STAGING_LIBDIR_NATIVE}`, so `4a1de5e8` is load-bearing for our layer too — through the `custom_lib_flags` route in `BUILD.gn.in` rather than clang's implicit search path. Both local workarounds for those are now deleted.

The two issues in my comment above are unchanged.

*(Claude-generated, Opus 5, from builds on this branch.)*

lhoward:
Tested on our device, working. The `GN_DEPS_SED_PATCHES` change I can also keep in a local `bbappend` if you don't want to apply it.

jwinarske:
Thanks @lhoward — all three confirmed and fixed on `3d081e46`. Two of them were mine.

**1. `GN_DEPS_SED_PATCHES` / duplicate `condition`** — I'd hit this independently on `qemuarm64` and fixed it by deleting the variable outright. **Your fix is better and I've taken it**, in `bf7591bc`:

```
'condition': 'download_android_deps and not (host_os == "linux" and host_cpu == "arm64")'
```

Deleting it parses, but as you note the JDK then lands in the cached source tarball — which matters more than I gave it credit for, since #776 makes that tarball the mirror artifact. Folding keeps the gate *and* upstream's arm64-host guard. Verified `not (host_os` occurs exactly once in 3.47.1's DEPS, so the anchor is unambiguous. Your point on quote-free anchors is right too.

**2. `GN_DEPS_SED_PATCHES` in no task signature** — correct, and a gap in my own `4960b178`: I named four of the six variables `lib/gn.py` reads and missed `GN_DEPS_SED_PATCHES` and `GN_EXTRA_CUSTOM_DEPS`. The former was worse than you describe — it isn't folded into the gclient config either, so it was missing from the **tarball cache key** as well as the signature. Editing it invalidated nothing anywhere, which is exactly the `0 tasks … all succeeded` you saw.

`58acceb5` adds both to `do_fetch[vardeps]` and puts the sed patches into the cache key. Sigdata now carries all six and its hash moves when they change.

**3. Retained partial sync unusable** — reproduced: `git remote add origin` exits **3** on a populated tree, so the 14GB is preserved and then rejected before gclient runs. `3d081e46` applies your `|| git remote set-url` (which also keeps the url correct if it changed between attempts). You're right that `4960b178` made this much easier to hit — `do_fetch` now re-runs on a config change and lands on an existing tree.

**On the rule 4 correction** — appreciated, and it resolves something I'd left ambiguous. I scoped the exemption to `CLANG_PATH` under `STAGING_DIR_NATIVE` precisely because it looked like a deliberate riscv64 deviation; your data confirms a stock glibc/aarch64 consumer doesn't trip rule 4 at all, so the exemption is riscv64-only in practice. That's the intended reading of `be7b7b72`.

Good to hear the device build is clean, and that you can drop the local workarounds.

No need to carry anything in a bbappend — all three are in the branch now.

jwinarske:
Need smoke tests on each architecture and green build test from @AhmedAdelWafdy7  on arm64 host prior to merging

lhoward:
Summary of where this stands from our side, superseding my earlier comment.

`core-image-xebros` (aarch64 / glibc / systemd, Flutter engine + a Swift embedder) builds clean on `22e165d6` and boots on target, with **no meta-flutter changes carried downstream**. Three of the four things we hit are fixed on the branch:

- `5f1b12b3` — the `GN_DEPS_SED_PATCHES` duplicate `condition` key
- `547b4936` — `GN_DEPS_SED_PATCHES` missing from the fetch signature and cache key
- `1a53888a` — retained partial syncs being unresumable

**One item outstanding: `packcmd` caches build output as source.**

`S` is both the sync path and the build directory, so after a successful build the tree holds `engine/src/out/<config>/` alongside sources. Since `1a53888a`, a re-fetch no longer fails on that populated tree — it succeeds — and `packcmd` tars it wholesale:

```python
ud.packcmd = f'tar -I "pbzip2 -p{bb_number_threads}" -cf {ud.localpath} ./; ...'
```

Measured here when `547b4936` changed the cache key and forced a refetch:

```
flutter-engine-3.47.1-<srcrev>.tar.bz2               2.2 GB
flutter-engine-3.47.1-<srcrev>-a9b3709eda82.tar.bz2  5.2 GB   <- 4.8 GB of engine/src/out
```

Before `1a53888a` this was a loud failure (`error: remote origin already exists`). Now it is silent, and the cached artifact is wrong rather than absent. Worth closing before `jw/gn-premirror` lands, since a poisoned tarball could reach a shared mirror.

Excluding the build output fixes it regardless of how the tree got dirty:

```
tar ... --exclude=./engine/src/out -cf {ud.localpath} ./
```

Related edge, not observed but reachable: an unpacked tarball containing `out/` hands ninja prebuilt targets, so `do_compile` could no-op and ship artifacts from an earlier build. Our last build ran all 13927 ninja steps, so it did not happen here.

To be clear about provenance — the incomplete suggestion that led to this was mine: I proposed making the remote setup idempotent without accounting for why the retained tree was dirty.

*(Claude-generated, Opus 5, from builds on this branch.)*

**Update:** we have moved to #781, since it targets `wrynose` and that is the branch we track. The outstanding `packcmd` item above applies there too — `lib/gn.py` is identical on `jw/wrynose-flutter-3.47.1`, and the whole tree delta between the two branch tips is CI workflows, `roll_meta_flutter.py` and the two `layer.conf` files. Nothing else differs, so a fix on either branch covers both.

Building clean against `87d8b93b` with no meta-flutter changes carried downstream.

AhmedAdelWafdy7:
Build passed on Arm64 host!

jwinarske:
Fixed in `c8412696` on this branch, and `c98cfae1` on #781.

Took the fix a bit wider than the suggested exclude, for three reasons:

**Per-recipe, not one path.** `--exclude=./engine/src/out` only covers flutter-engine. dart-sdk builds in `sdk/out`, libwebrtc in `src/out` (its sync directory is `gclientdir`, not `S`), and pdfium builds outside the sync directory entirely and needs nothing. So it's `GN_PACK_EXCLUDES`, set per recipe. Explicit list rather than a glob on any directory named `out` — a pattern that matched a source directory would drop it silently, which is the failure we're fixing.

**In the signature and the cache key.** Changing what's excluded changes what the tarball holds, so it's in `do_fetch[vardeps]` and in `config_hash`. Otherwise we'd have rebuilt the trap `547b4936` just removed.

**Cleared before the sync, not just filtered while packing.** The same paths are removed before gclient runs, so a retained partial tree goes in clean rather than being filtered on the way out. That also closes the ninja edge you flagged.

On provenance — the idempotent remote setup was yours to suggest, mine to land without asking why the retained tree was dirty. Fair trade.

One correction: #776 is closed. Premirrors already worked — `uri_replace()` derives the mirror filename from `basename(ud.localpath)` whenever `supports_checksum()` is true, which it is by default, so `mirrortarballs` added nothing and the `supports_checksum() → False` I'd paired it with would have broken that path on its own. Which sharpens your point rather than softening it: a poisoned tarball can reach a shared mirror **today**, not after some future PR.

landing order: `master` -> `wrynose` -> `scarthgap` -> `kirkstone` -> `dunfell`.  When master is green and validated I'll move to `wrynose`.  Yes meta-flutter is labor and machine intensive.

Issue #271, opened 2023-03-20
reproducible engine build


jwinarske:
Requirement is to support a reproducible engine library.

This means the file can be compared to prior build, and there is not difference.

The solution revolves around removal of .git folders and any processes dependent on the .git folder.

## Round 1
```
gclient sync --revision ada363ee93b17cfe31587b5102679885cb40837e --no-history --shallow
find . -name .git -type d -print0|xargs -0 rm -r --
./flutter/tools/gn --unoptimized --no-goma
```
It fails with 
```
Generating GN files in: out/host_debug_unopt
ERROR at //flutter/shell/version/version.gni:17:7: Script returned non-zero exit code.
      exec_script("//flutter/build/git_revision.py",
      ^----------
Current dir: /mnt/raid10/flutter-engine/src/out/host_debug_unopt/
Command: vpython3 /mnt/raid10/flutter-engine/src/flutter/build/git_revision.py --repository /mnt/raid10/flutter-engine/src/flutter
Returned 1.
stderr:

fatal: not a git repository (or any parent up to mount point /mnt)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
Traceback (most recent call last):
  File "/mnt/raid10/flutter-engine/src/flutter/build/git_revision.py", line 58, in <module>
    sys.exit(main())
  File "/mnt/raid10/flutter-engine/src/flutter/build/git_revision.py", line 51, in main
    version = get_repository_version(repository)
  File "/mnt/raid10/flutter-engine/src/flutter/build/git_revision.py", line 28, in get_repository_version
    version = subprocess.check_output([
  File "/home/joel/depot_tools/.cipd_bin/3.8/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/home/joel/depot_tools/.cipd_bin/3.8/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', '-C', '/mnt/raid10/flutter-engine/src/flutter', 'rev-parse', 'HEAD']' returned non-zero exit status 128.

See //flutter/tools/path_ops/BUILD.gn:5:1: whence it was imported.
import("//flutter/shell/version/version.gni")
^-------------------------------------------
See //flutter/BUILD.gn:93:9: which caused the file to be included.
[...]

PR #776, opened 2026-08-21
gn-fetcher: support premirrors

jwinarske:
The gn fetcher never told bitbake that the tree it packs is a mirrorable artifact, so `PREMIRRORS`/`MIRRORS` were only matched against the `gn://` url itself. In practice they didn't apply, `BB_FETCH_PREMIRRORONLY` couldn't be satisfied, and a build without access to the upstream repositories couldn't fetch the engine at all.

### Change

- **`ud.mirrortarballs`** — set to the packed tarball name, so bitbake builds mirror uris against that filename rather than the `gn://` url. Unlike the git fetcher there's no separate clone dir to mirror alongside it, so nothing for `BB_GENERATE_MIRROR_TARBALLS` to gate; the tarball is always written.
- **`supports_checksum() → False`** — the tarball is repacked from a gclient sync and isn't bit-identical between runs, because the tree keeps the `.git` directories the flutter build needs for `git rev-parse`. Same position as the other VCS fetchers.
- **`try_premirror()`** — follows the git fetcher: always consult premirrors under `BB_FETCH_PREMIRRORONLY` or an untrusted network, otherwise only when the tarball isn't already in `DL_DIR`.

### Verified end to end

With a local mirror seeded and **both `BB_NO_NETWORK=1` and `BB_FETCH_PREMIRRORONLY=1`**:

```
NOTE: recipe flutter-engine-3.47.1-r0: task do_fetch: Succeeded
downloads/flutter-engine-3.47.1-<srcrev>-<confighash>.tar.bz2 -> <mirror>/flutter/...
```

Before the change, the identical configuration fell through to `Trying Upstream` and failed with `NetworkAccess('gn://github.com/flutter/flutter.git', ...)`.

### Writing the mirror entry

Two things worth knowing, both found while testing:

1. oe-core's mirror sanity check (`meta/lib/oe/sanity.py`) has a **hardcoded protocol list** predating out-of-tree fetchers, without `gn` — so naming `gn://` explicitly warns. Use the generic `.*://.*` pattern, as OE's own default `PREMIRRORS` does.
2. The mirror layout follows the url path, so the tarball for `gn://github.com/flutter/flutter.git` belongs under `<mirror>/flutter/`.

### Relationship to #269

The premirror groundwork was proposed in #269 by @camaronut; this reimplements it against the current fetcher, which has since gained `gclient_config`, `EXTRA_GN_SYNC` and the config hash from #773.

The **reproducibility half of #269 is deliberately left out**:
- The tarball can't be made reproducible while the `.git` directories have to stay — the premise doesn't hold (see #271).
- The `.packages` and `package_config.json` files it rewrote **no longer exist** in a current engine tree (checked a 3.47.1 tree: zero of each). `.packages` has been deprecated in Dart for years.
- `SOURCE_DATE_EPOCH` isn't meaningful during `do_fetch` — `do_deploy_source_date_epoch` is `addtask ... after do_patch`, so `urldata_init` would get the fallback, not a real epoch.
- Rewriting files during fetch also mutates a tree that's now cached under a key that knows nothing about those edits.

**Depends on #773** (based on `jw/flutter-3.47.1-ihs-v3`), which rewrites the same `ud.localfile` line. GitHub will retarget to `master` when #773 merges.

### DIFF
--- a/lib/gn.py
+++ b/lib/gn.py
@@ -18,6 +18,7 @@ import urllib
 from   bb.fetch2 import FetchMethod
 from   bb.fetch2 import FetchError
 from   bb.fetch2 import UnpackError
+from   bb.fetch2 import trusted_network
 from   bb.fetch2 import logger
 from   bb.fetch2 import runfetchcmd
 from   bb.fetch2 import subprocess_setup
@@ -97,6 +98,17 @@ class GN(FetchMethod):
 
         ud.localfile = d.getVar("PN") + '-' + d.getVar("PV") + "-" + srcrev + "-" + config_hash + ".tar.bz2"
         ud.localpath = os.path.join(d.getVar("WORKDIR"), ud.localfile)
+
+        # The packed tree is the only artifact this fetcher produces, so it is
+        # also the thing worth mirroring. Naming it here lets bitbake build
+        # mirror URIs against the tarball filename rather than against the gn://
+        # url, which is what makes PREMIRRORS/MIRRORS usable -- and with it
+        # BB_FETCH_PREMIRRORONLY, so a build with no access to the upstream
+        # repositories can still fetch. Unlike the git fetcher there is no
+        # separate clone directory to mirror alongside it, so there is nothing
+        # for BB_GENERATE_MIRROR_TARBALLS to gate: the tarball is always written.
+        ud.fullmirror = os.path.join(d.getVar("DL_DIR"), ud.localfile)
+        ud.mirrortarballs = [ud.localfile]
 
         ud.trying_to_fetch_with_gclient = False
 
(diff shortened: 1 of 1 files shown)


jwinarske:
Closing this — **the premise was wrong, and the change is a no-op.**

I claimed premirrors "did not apply" to gn recipes and that a `BB_FETCH_PREMIRRORONLY` build could not fetch. I validated the *positive* case (mirror seeded, `BB_NO_NETWORK=1`, `do_fetch` succeeded) but never ran the control, so I never established that the change was what made it work. It isn't.

### The control

`flutter-engine` is a poor test target — it has a `git://` entry with no offline mirror artifact, which confounds the comparison. `dart-sdk` has only `gn://` and `file://`:

| | `do_fetch` | DL_DIR | went upstream | verdict |
|---|---|---|---|---|
| **without this PR** | 0 | present | no | **works** |
| **with this PR** | 0 | present | no | works |

Premirrors already work for gn recipes on master.

### Why

`uri_replace()` in `bb/fetch2/__init__.py` derives the mirror filename two ways:

```python
if uri_decoded[0] != uri_replace_decoded[0] and mirrortarball:
    basename = os.path.basename(mirrortarball)          # what this PR adds
elif ud.localpath and ud.method.supports_checksum(ud):
    basename = os.path.basename(ud.localpath)           # what already worked
```

The `elif` was already yielding the tarball name, since the default `supports_checksum()` returns True for a plain file in `DL_DIR`.

Worse, this PR's two halves cancel out: `supports_checksum() → False` **disables** the `elif` that was doing the work, and `mirrortarballs` then restores the same basename through the `if`. Either half alone would break premirroring — `supports_checksum → False` on its own is strictly harmful. So it replaces a working mechanism with an equivalent but more fragile one.

### What was actually worth having

Not code — documentation. Premirroring a gn recipe works today, with two non-obvious details, both found while testing this:

1. Use the generic `.*://.*` pattern. oe-core's mirror sanity check (`meta/lib/oe/sanity.py`) carries a hardcoded protocol list that predates out-of-tree fetchers and has no `gn`, so naming `gn://` explicitly warns.
2. The mirror layout follows the url path — the tarball for `gn://github.com/dart-lang/sdk.git` goes at `<mirror>/dart-lang/`, not the mirror root.

I'll put that in the layer README instead.

This also means #269's premirror half was solving a problem that no longer exists in the current fetcher, so closing that in favour of this PR was itself the wrong call — apologies to @camaronut. The reproducibility half of #269 remains blocked on #271 as described there.

PR #778, opened 2026-08-21
ci: run runner-maintenance without a container

jwinarske:
Fixes the first dispatch of #777, which failed before any step ran:

```
Error: statfs /mnt/raid10/actions-runner/_work/_actions: no such file or directory
##[error]Exit code 125 returned from process: file name '/usr/bin/docker'
```

### Cause

A container job bind-mounts `_work/_actions`, which the runner only materialises once a workflow has used an action. Every other workflow here opens with `actions/checkout@v4`, so it's normally present — `runner-maintenance` has no `uses:` step and hit the gap. Confirmed `_work/_actions` is absent on the runner right now.

### Fix

Adding a checkout would paper over it, but the container earns nothing here: this job needs no build image, no sysroots, no toolchain — only `du`, `find` and `rm`. Dropping it removes the image pull *and* the bind-mounts along with the failure, and it still runs as uid 1001, which owns the tree.

Also references the build tree through `$GITHUB_WORKSPACE` rather than a relative path, since without a checkout nothing guarantees the working directory.

Everything else is unchanged: still defaults to `report`, still requires `confirm=wipe` for the destructive path, path still fixed in the workflow rather than an input.

Unblocks the `build/tmp` cleanup that `qemuarm`/`qemuarm64` on #773 need.

### DIFF
--- a/.github/workflows/runner-maintenance.yml
+++ b/.github/workflows/runner-maintenance.yml
@@ -35,28 +35,18 @@ jobs:
 
   maintenance:
 
+    # Deliberately no container: this needs no build image, and a container
+    # job bind-mounts _work/_actions, which only exists once a workflow has
+    # used an action. Running natively on the runner avoids that entirely and
+    # is already the uid that owns the tree.
     runs-on: [self-hosted, linux]
-
-    container:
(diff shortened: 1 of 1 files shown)


jwinarske:
Superseded by the combined CI PR — folded in there to save a build round.

PR #781, opened 2026-08-22
wrynose: roll to Flutter 3.47.1 and add ivi-homescreen v3

jwinarske:
Backport of #773 to `wrynose`. All 12 commits cherry-picked **clean, zero conflicts** — this branch is 2 files from master.

Plus one branch-specific commit.

### CI: every machine, on this branch's own tree

The workflows here were master's unchanged — same workflow *names*, and `YOCTO_BRANCH: master`, so they cloned a **blacksail** openembedded-core and built in `master-linux-dummy` with master's sstate. This branch has therefore never actually been verified against wrynose, and every release branch carrying these files would contend for one build tree.

Replaced with one workflow covering **all four machines** as a matrix — `qemux86-64`, `qemuarm64`, `qemuriscv64`, `qemuarm` — against openembedded-core **wrynose**, in this branch's own tree and sstate.

**Dispatch rather than per-PR is a disk constraint, not a coverage one.** A populated `tmp` measures ~**106G** against **264G free**, so master plus four release branches can't each keep one. The runner is sequential, so the matrix queues one machine at a time against a single tree — which is how master's `tmp` comes to hold all four already. master's tree is never touched, so its per-PR builds stay fast, and any *other* release branch's `tmp` is cleared first, holding the release branches to one populated tree between them.

The fetch step carries master's fixes (#774): every cloned layer listed in the `rm -rf`, and clone pids waited on individually.

App steps are skipped for `qemuarm` — `flutter build bundle` has no `linux-arm` target platform, so those recipes don't parse for a 32-bit machine. Same as master.

### Layer series

`LAYERSERIES_COMPAT` stays `"whinlatter wrynose"` rather than the `"wrynose blacksail"` #773 sets on master. That's only correct now that the build uses a wrynose openembedded-core.

### oe-core variable rename

`DISTRO_FEATURES_BACKFILL_CONSIDERED` → `DISTRO_FEATURES_OPTED_OUT`; oe-core errors on the old name. Went to master separately in #780.

### Validation

**Not yet built.** Once merged, dispatch `wrynose-build` — it covers all four machines in one run. Depends on #773 landing first so the two don't diverge.


jwinarske:
CI failures are not related to meta-flutter

PR #788, opened 2026-08-23
flutter-app: pub workspace support, and the flutter_scene smoke_render recipe

jwinarske:
Groundwork for building applications that are **pub workspace members**. Landing it separately from the recipe that needs it, since it touches code shared by **191 app recipes**.

### Problem

A workspace member can't be resolved from its own directory — the lockfile lives at the workspace root and pub refuses to operate in the member. Every pub step ran in `${S}/${FLUTTER_APPLICATION_PATH}`, so such an app couldn't be built at all.

### Change

`FLUTTER_PUB_ROOT` — the directory pub resolves from, defaulting to the application:

```
FLUTTER_PUB_ROOT ??= "${FLUTTER_APPLICATION_PATH}"
```

| moves to pub root | stays on the application |
|---|---|
| `flutter config`, `flutter clean` | pub cache **archive name** |
| lockfile handling | `PUB_CACHE_EXTRA_ARCHIVE_CMD` hook |
| both `pub get` invocations | |
| copy + restore of the resolution tree | |

The resolution tree moves because pub writes `.dart_tool` **where it ran**, not in the member — archive and restore have to stay symmetric or a workspace build loses its resolution.

Keying the cache on the *application* is deliberate: one pub root can host several applications, so keying on the root would have them share an entry.

### Verified, not assumed

**Default is a no-op** — `bitbake -e` on three existing recipes, including a nested path:

```
jwinarske-bluez-native-flutter-ble-scanner   app='example/flutter_ble_scanner'          pub=same
flatpak-minimal-appstream-dart-flathub-catalog app='example/flathub_catalog'            pub=same
toyota-connected-…-camera-linux-camera-example app='packages/camera/camera_linux/example' pub=same
```

Full tree parses (5366 recipes).

### One real cost

`do_archive_pub_cache`'s signature changes, because the task body reads one more variable — measured on an unmodified recipe:

```
without: ef76fa4fd698f979
with:    5d1813d67f1e3066
```

So **every app recipe re-archives its pub cache once**. That's a re-resolve and repack, not a rebuild, and it's unavoidable for any edit to a shared task — but it's a real cost on the next build after this merges.

### DIFF
--- a/conf/include/common.inc
+++ b/conf/include/common.inc
@@ -10,6 +10,13 @@ DEPENDS += " \
     "
 
 FLUTTER_APPLICATION_PATH ??= ""
+
+# Directory pub resolves from, relative to S. Defaults to the application, which
+# is where a standalone package resolves. A pub workspace member resolves from
+# the workspace root instead -- its own directory has no lockfile and pub
+# refuses to operate there -- so such a recipe points this at the root and
+# leaves FLUTTER_APPLICATION_PATH on the member.
+FLUTTER_PUB_ROOT ??= "${FLUTTER_APPLICATION_PATH}"
 FLUTTER_PREBUILD_CMD ??= ""
 
 PUB_CACHE_EXTRA_ARCHIVE_CMD ??= ""
@@ -186,6 +193,10 @@ python do_archive_pub_cache() {
                 return
 
     app_root = os.path.join(d.getVar("S"), d.getVar("FLUTTER_APPLICATION_PATH"))
+    # pub runs here; for a workspace member this is the workspace root rather
+    # than the application. The cache is still keyed on the application, since
+    # that identifies what is being built -- one pub root can host several.
+    pub_root = os.path.join(d.getVar("S"), d.getVar("FLUTTER_PUB_ROOT"))
     localfile = get_pub_cache_archive_name(d, app_root)
     localpath = os.path.join(d.getVar("DL_DIR"), localfile)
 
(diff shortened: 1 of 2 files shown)


jwinarske:
Added the recipe that motivated this, so both are reviewable together — `c4a2b903`.

### `bdero-flutter-scene-smoke-render_1.0.0.bb`

```
FLUTTER_PUB_ROOT = ""                              # workspace root
FLUTTER_APPLICATION_PATH = "examples/smoke_render" # the member
PUBSPEC_APPNAME = "smoke_render"
inherit flutter-app-native
```

Chosen as the **first** flutter_scene example deliberately: shader and material hooks, but no Rust, no audio, no workspace-only native packages. `flutter_app` needs a cargo toolchain and vendored crates, which is a separate piece of work.

**Why the workspace handling isn't merely mechanical.** The app depends on `flutter_scene: ^0.23.0` *by version*, and the repo carries its own `packages/flutter_scene` at **0.23.0**. Workspace resolution binds to the local copy — what the example is written against. Resolving the member standalone would take the *published* 0.23.0: same version number, different code, no error. That's what ruled out the de-workspacing alternative.

`flutter-app-native` rather than `flutter-app`: the hook drives no CMake itself, but `flutter_scene` depends on `code_assets` as well as `data_assets`, so a hook may emit native code. Superset, otherwise equivalent.

### Verified

Recipe parses; `FLUTTER_PUB_ROOT` resolves empty and `FLUTTER_APPLICATION_PATH` to `examples/smoke_render`; tree goes 5366 → **5367** recipes.

### Not verified

**Not built.** And one upstream gap worth knowing before this ships: flutter_scene commits **no `pubspec.lock`**, at the root or in the member. So `--enforce-lockfile` has nothing to enforce, and dependency versions are whatever pub resolves at first fetch rather than something upstream pinned. Reproducible after that first fetch via the pub-cache archive, but not pinned by upstream intent.

Runtime prerequisites are proven separately — built and ran this example on ivi-homescreen v3 (`wayland-vulkan`) on this host, with `--engine-arg=--enable-flutter-gpu` reaching `command_line_argv` and Impeller/Vulkan initialising.

PR #798, opened 2026-08-24
flutter-app: key the pub cache skip on the source, not a stale marker

### LINKED ISSUE #653: `do_archive_pub_cache` not always re-run [SRCREV = "${AUTOINC}"]
We're having an issue where the `do_archive_pub_cache` task is not always being re-run.

We have a recipe for an app that has two versions: a stable version with a set version number and a `git+` version that builds the latest commit. It's this git version that isn't working properly. If we step the commit forwards (or use `${AUTOINC}`) then `do_archive_pub_cache` does nothing. The logs show the following:

```
DEBUG: Executing python function extend_recipe_sysroot
NOTE: Direct dependencies are ['/home/ed/kas/build/../oe-core/meta/recipes-devtools/quilt/quilt-native_0.67.bb:do_populate_sysroot', 'virtual:native:/home/ed/kas/build/../meta-flutter/recipes-graphics/flutter-sdk/flutter-sdk_git.bb:do_populate_sysroot', 'virtual:native:/home/ed/kas/build/../oe-core/meta/recipes-devtools/patch/patch_2.7.6.bb:do_populate_sysroot', 'virtual:native:/home/ed/kas/build/../oe-core/meta/recipes-extended/pbzip2/pbzip2_1.1.13.bb:do_populate_sysroot']
NOTE: Installed into sysroot: []
NOTE: Skipping as already exists in sysroot: ['gettext-minimal-native', 'cmake-native', 'libtool-native', 'm4-native', 'quilt-native', 'texinfo-dummy-native', 'flutter-sdk-native', 'openssl-native', 'dbus-native', 'expat-native', 'gettext-native', 'glib-2.0-native', 'libxml2-native', 'ncurses-native', 'util-linux-libuuid-native', 'util-linux-native', 'zlib-native', 'flex-native', 'gnu-config-native', 'libedit-native', 'make-native', 'patch-native', 'perl-native', 'python3-build-native', 'python3-flit-core-native', 'python3-installer-native', 'python3-packaging-native', 'python3-pyproject-hooks-native', 'python3-setuptools-native', 'python3-wheel-native', 'python3-native', 'bzip2-native', 'libnsl2-native', 'libtirpc-native', 'pbzip2-native', 'unzip-native', 'xz-native', 'zstd-native', 'gdk-pixbuf-native', 'gobject-introspection-native', 'gtk+3-native', 'cairo-native', 'fontconfig-native', 'freetype-native', 'harfbuzz-native', 'libjpeg-turbo-native', 'pango-native', 'libpthread-stubs-native', 'libx11-native', 'libxau-native', 'libxcb-native', 'libxcomposite-native', 'libxcursor-native', 'libxdamage-native', 'libxdmcp-native', 'libxext-native', 'libxfixes-native', 'libxft-native', 'libxi-native', 'libxrandr-native', 'libxrender-native', 'libxtst-native', 'pixman-native', 'xtrans-native', 'xcb-proto-native', 'xorgproto-native', 'util-macros-native', 'libpng-native', 'at-spi2-core-native', 'attr-native', 'ca-certificates-native', 'curl-native', 'fribidi-native', 'gdbm-native', 'icu-native', 'itstool-native', 'libcap-ng-native', 'libffi-native', 'libpcre2-native', 'shared-mime-info-native', 'sqlite3-native']
DEBUG: Python function extend_recipe_sysroot finished
DEBUG: Executing python function do_archive_pub_cache
DEBUG: Python function do_archive_pub_cache finished
```

We're on scarthgap, which changed the way package versions are done. Our recipe contains the following:

```
SRCREV = "d9c2b71530102d5e2c6c4b326e76a0bbb5b0cd3d"
PV = "git"
```

Which should set `PKGV` to be `git+d9c2b71530102...`

So for some reason `do_archive_pub_cache` seems to depend on `PV` (which is static in scarthgap) but not `PKGV` or even the result of `do_patch` for when the source code changes.

I've been doing some digging at my end and am still testing the effect of adding `do_archive_pub_cache[deptask] = "do_patch"` to the class

jwinarske:
Fixes #653 — `do_archive_pub_cache` not re-running when `SRCREV` moves.

### The bug

The task opens with a guard driven by a marker file written by an *earlier* build:

```python
pub_cache_archive_file = os.path.join(workspace, 'PUB_CACHE_ARCHIVE')
if os.path.exists(pub_cache_archive_file):
    with open(pub_cache_archive_file, 'r') as f:
        localfile = f.read()
        if os.path.exists(localfile):
            return          # ← returns before the new md5 is computed
```

Bump `SRCREV` and the source changes, so the archive name *would* change — its md5 is taken over the application directory. But the guard returns before `get_pub_cache_archive_name()` is ever called. The marker still names the previous archive, that archive is still in `DL_DIR`, so the task no-ops and the app builds against the previous run's pub cache.

That is exactly the log in #653: the task executes and finishes with nothing between.

Worth being explicit, since this came up in the issue thread: the naming logic was never wrong. The md5 does change with the source. The problem is that it is computed *after* the decision that depends on it.

### Why CI never caught it

`WORKDIR` has to survive between builds for the marker to go stale. `ci-build.inc` does `INHERIT += "rm_work"`, which wipes it — so this reproduces on a developer tree and never here.

### Second, latent bug

`localfile` is a bare filename, and `os.path.exists(localfile)` resolved it against the cwd. It matched only because `do_archive_pub_cache[dirs] = "${WORKDIR} ${DL_DIR}"` leaves bitbake chdir'd into `DL_DIR`. Correct by accident — reordering that flag would have silently disabled the skip. The fix tests `localpath`.

### The change

Name the archive first, then test for that file:

```python
localfile = get_pub_cache_archive_name(d, app_root)
localpath = os.path.join(d.getVar("DL_DIR"), localfile)
if os.path.exists(localpath):
    return
```

The skip is now decided by the source rather than by a record of what was built last time. `do_restore_pub_cache` still reads the marker and `get_pub_cache_archive_name()` still writes it, so nothing downstream changes, and an unchanged source still produces the same name and still skips.

### Verification

The mechanism, exercised against the layer's own `md5_dir` with two source trees differing by one line:

```
md5 before SRCREV bump : 6c0fbe182c3384617ba6b1c42ba0df56
md5 after  SRCREV bump : d0c5675bcfe8e4d770c09d11cbeabbdc
archive name would change: True

guard: os.path.exists(flutter-pub-cache-myapp-6c0fbe182c...) -> True
=> task returns early, never computes d0c5675b..., archive is never rebuilt
```

All `def`s and `python` tasks in `common.inc` parse cleanly after the edit.

**Not reproduced in a full build.** This is a unit-level demonstration of the mechanism plus code reading, not a bitbake run with a bumped `SRCREV` on a non-`rm_work` tree. That is the test worth doing before merge if you want belt and braces — the reporter's own workaround (putting `SRCREV` back into `PV`) also suggests it, since that changes `WORKDIR` and so sidesteps the stale marker.

### DIFF
--- a/conf/include/common.inc
+++ b/conf/include/common.inc
@@ -184,21 +184,44 @@ python do_archive_pub_cache() {
     import shutil
     from bb.fetch2 import FetchError
 
-    workspace = d.getVar('WORKDIR')
-    pub_cache_archive_file = os.path.join(workspace, 'PUB_CACHE_ARCHIVE')
-    if os.path.exists(pub_cache_archive_file):
-        with open(pub_cache_archive_file, 'r') as f:
-            localfile = f.read()
-            if os.path.exists(localfile):
-                return
-
     app_root = os.path.join(d.getVar("S"), d.getVar("FLUTTER_APPLICATION_PATH"))
     # pub runs here; for a workspace member this is the workspace root rather
     # than the application. The cache is still keyed on the application, since
     # that identifies what is being built -- one pub root can host several.
     pub_root = os.path.join(d.getVar("S"), d.getVar("FLUTTER_PUB_ROOT"))
+
+    # Skipping needs two things to be true, and the original code checked only
+    # the first: the marker has to name what this WORKDIR already built, and
+    # that archive has to still be there.
+    #
+    # Trusting the marker alone was the bug in #653. It records a name from an
+    # earlier build, so a bumped SRCREV returned here before the md5 of the new
+    # source was ever computed, and the app then built against the previous
+    # run's pub cache. Recomputing the name first is what detects that.
+    #
+    # Trusting the archive alone is wrong the other way. DL_DIR outlives
+    # WORKDIR -- it is a shared, persistent mount in CI, while rm_work and
+    # cleansstate wipe WORKDIR every build -- so the archive is routinely
+    # present when this recipe has done no work at all. Returning there skips
+    # the pub cache population below, and do_compile then fails on
+    # "flutter config --no-analytics". The name carries PN, the source md5 and
+    # the SDK tag but no architecture, so one machine populating DL_DIR was
+    # enough to break the rest.
+    marker = os.path.join(d.getVar('WORKDIR'), 'PUB_CACHE_ARCHIVE')
+    previous = None
+    if os.path.exists(marker):
+        with open(marker, 'r') as f:
+            previous = f.read().strip()
+
+    # Writes the marker as a side effect, so it has to come after the read.
     localfile = get_pub_cache_archive_name(d, app_root)
     localpath = os.path.join(d.getVar("DL_DIR"), localfile)
+
+    # localpath, not the bare name: the original resolved against the cwd and
+    # only worked because do_archive_pub_cache[dirs] leaves bitbake in DL_DIR.
+    if previous == localfile and os.path.exists(localpath):
+        bb.note(f'pub cache already archived for this source: {localfile}')
+        return
 
     pub_cache = d.getVar("PUB_CACHE")
     flutter_sdk = os.path.join(d.getVar("STAGING_DATADIR_NATIVE"), 'flutter/sdk')


jwinarske:
The first version of this regressed CI, on `qemuarm64` and `qemuriscv64`:

```
ERROR: ...camera-linux-camera-example do_compile: flutter config --no-analytics failed: 1
```

That was my bug, and it is worth recording because it is the mirror image of the one being fixed.

### What went wrong

The original guard required **two** things to skip: the marker existed *and* the file it named existed. My version required only that the recomputed archive was present in `DL_DIR`. Those differ in exactly CI's situation:

| | marker (`WORKDIR`) | archive (`DL_DIR`) | original | first fix |
|---|---|---|---|---|
| CI rebuild | absent — `cleansstate`/`rm_work` wipe it | present — persistent mount | does the work | **returns early** |

Returning there skips the pub cache population that follows, so `do_compile` fails later. And because the archive name is `PN` + source md5 + SDK tag with **no architecture**, all four machines share one name in a shared `DL_DIR` — whichever ran first created it and passed (`qemuarm`), and the rest found it and skipped.

### The corrected guard

Require both conditions:

```python
marker = os.path.join(d.getVar('WORKDIR'), 'PUB_CACHE_ARCHIVE')
previous = open(marker).read().strip() if os.path.exists(marker) else None

localfile = get_pub_cache_archive_name(d, app_root)   # writes the marker
localpath = os.path.join(d.getVar("DL_DIR"), localfile)

if previous == localfile and os.path.exists(localpath):
    return
```

Behaviour across the cases that matter, with each variant failing exactly one:

| case | original | first fix | corrected |
|---|---|---|---|
| CI rebuild: `WORKDIR` wiped, `DL_DIR` persists | skip=False ✓ | **skip=True ✗** | False ✓ |
| same tree, unchanged source | True ✓ | True ✓ | True ✓ |
| bumped SRCREV, old archive still present | **True ✗** | False ✓ | False ✓ |
| first ever build | False ✓ | False ✓ | False ✓ |

Cache hit preserved for an unchanged rebuild, work redone when the source moves, and work always done when the tree is fresh regardless of `DL_DIR`.

### Note

This is precisely the failure mode the previous PR body flagged as untested. Reasoning from the pipeline found the staleness bug but missed that `DL_DIR` outlives `WORKDIR` — which only a real build would surface, and did.

Issue #819, opened 2026-08-29
AOT images embed TMPDIR via the Dart plugin registrant URI (buildpaths QA)


jwinarske:
Tracking issue for porting the fix from [toyota-connected/emb_cli#188](https://github.com/toyota-connected/emb_cli/pull/188) (issue [#186](https://github.com/toyota-connected/emb_cli/issues/186)) into meta-flutter.

This resolves the **"AOT generation"** checkbox of #566, whose reported symptom is:

```
WARNING: mogol-flutter-secure-storage-flutter-secure-storage-example-git-r0 do_package_qa:
QA Issue: File /usr/share/flutter/.../release/lib/libapp.so in package ...
contains reference to TMPDIR [buildpaths]
```

#566 attributes this to [dart-lang/sdk#55282](https://github.com/dart-lang/sdk/issues/55282). That upstream issue is real, but it is not what produces *this* warning, and this one we can fix ourselves.

## What actually embeds TMPDIR

`conf/include/common.inc:506-510` names the generated plugin registrant to the frontend server by absolute path, twice:

```python
dart_plugin_registrant_file = f'{source_dir}/{flutter_application_path}/.dart_tool/flutter_build/dart_plugin_registrant.dart'
if os.path.isfile(dart_plugin_registrant_file):
    flutter_source_file    = f'--source file://{dart_plugin_registrant_file}'
    flutter_source_package = '--source package:flutter/src/dart_plugin_registrant.dart'
    flutter_source_defines = f'-Dflutter.dart_plugin_registrant=file://{dart_plugin_registrant_file}'
```

`source_dir` is under `WORKDIR`, hence under `TMPDIR`. That URI becomes the generated library's **import URI**, which the AOT image retains verbatim.

Everything else in the image is already clean: the app and the SDK are reached through `--packages`, so their libraries carry `package:` URIs, not file paths. The registrant is the one library with no package URI, so it is the one that leaks — and it is the *only* one.

Measured, by building a plugin app with meta-flutter's exact gen_snapshot flag set (`--strip`, no `--obfuscate`):

| | absolute paths in `libapp.so` |
|---|---|
| today | **1** — the registrant |
| with the fix | **0** |

Confirmed independently on a release build of `flutter_secure_storage/example`, the very app named in #566's warning: its `libapp.so` contains exactly one absolute path, and it is the registrant.

[...]

Issue #566, opened 2024-09-10
[scarthgap] reproducible builds


jwinarske:
This is an umbrella issue for reproducible build issues on `scarthgap`

- [ ] AOT generation.  https://github.com/dart-lang/sdk/issues/55282
- [ ] flutter-sdk-native in Yocto Toolchain

cc @g-scott-murray

camaronut:
Hi Joel,

I've been on a 2+ year tour of the bowels of Android, hopefully my employer will give me more time to come back to this now that the trip is winding itself down.

A colleague and I were updating our sources from Dunfell to Scarthgap.  Our company's build flow is typically "do Yocto release of BSP and SDK, give to app developers, repeat".  The app developers then take that SDK and build a flutter front-end and a C++ back-end app, linked together with protobuf.  The app is provisioned to the device separate from the Yocto generated filesystem.  App developers use the pub-cache folder that exists inside the Yocto generated SDK.

We have a .bbappend file in our layer for flutter-sdk_git that adds a step to attempt to bolt-down protoc:
```
do_compile[network] = "1"

do_compile:append() {
	# Use a cache path inside the SDK staging area, not $HOME
	export PUB_CACHE="${S}/.pub-cache"
	# The 25.0.0 is intentional - it bolts down the version
	${S}/bin/dart pub global activate protoc_plugin 25.0.0
}
```
We may move this to do_unpack, that's still ongoing.

We were thinking about this today, the SDK's .pub-cache folder may not be reproducible... should that be added to the list here?

We also found this from googling:
https://pub.dev/packages/pub2yocto

It seems pub2yocto could perhaps be used ahead-of-time to help write the SRC_URIs of the dart packages/plugins from a given dart/flutter setup, and that could perhaps be used to feed an SDK build recipe.  Unfortunately this also sounds like a maintenance burden to make updates.

If you have any creative ideas on how to improve the pub-cache handling inside the SDK, we're all ears.
Thanks!

jwinarske:
The **AOT generation** item here splits into two independent leaks. Both are now tracked:

[...]

PR #821, opened 2026-08-29
master: keep TMPDIR out of the AOT image

jwinarske:
Port of #820 (wrynose, green on all four machines) for #819. Resolves the **AOT generation** item of #566.

The diff is byte-identical to the wrynose one.

## The leak

```
QA Issue: File /usr/share/flutter/.../release/lib/libapp.so in package ...
contains reference to TMPDIR [buildpaths]
```

`common.inc` named the generated plugin registrant by absolute path, twice — as `--source` and as the `-Dflutter.dart_plugin_registrant` define. `source_dir` is under `WORKDIR`, hence under `TMPDIR`, and that path becomes the generated library's **import URI**, which the AOT image keeps verbatim.

The registrant is the *only* library an app compiles that has no `package:` URI — everything else is reached through `--packages`. Measured by building a plugin app with meta-flutter's exact gen_snapshot flag set (`--strip`, no `--obfuscate`):

| | absolute paths in `libapp.so` |
|---|---|
| before | **1** — the registrant |
| after | **0** |

Confirmed independently against a release build of `flutter_secure_storage/example`, the app in #566's warning.

## Why not just make it relative

The frontend canonicalizes a bare path back to an absolute `file:` URI; the relative form then appears in the image *alongside* the absolute one. The URI has to be in a scheme with no absolute form, so the app is declared as a filesystem root and the registrant named against it:

```
org-dartlang-root:///.dart_tool/flutter_build/dart_plugin_registrant.dart
```

Same shape `flutter build` produces via `toMultiRootPath()`, but only when given roots — a desktop AOT build has none and falls back to the absolute URI, so it has to be set here.

## Registration still works

Verified in [emb_cli#188](https://github.com/toyota-connected/emb_cli/pull/188), which builds the AOT through the same frontend/gen_snapshot sequence: statically the string sets are identical bar the registrant's URI-derived library hash, and at runtime an app whose `main()` awaits `SharedPreferences.getInstance()` — Dart-only, reachable *only* through the registrant — reports `registrant OK, readback=42` both before and after.

wrynose CI additionally built `flatpak-minimal-flatpak-dart-flutter-remote-manager`, whose registrant carries five real `registerWith()` calls including `path_provider_linux`, so this code path executed in a real Yocto build and compiled green.

## Scope

Profile builds still carry source URIs, from `--track-widget-creation`. Release is what ships, and release is clean.

### DIFF
--- a/conf/include/common.inc
+++ b/conf/include/common.inc
@@ -523,11 +523,27 @@ def build_app(d, source_dir, flutter_sdk, pubspec_appname, flutter_application_p
             flutter_source_file = ''
             flutter_source_package = ''
             flutter_source_defines = ''
-            dart_plugin_registrant_file = f'{source_dir}/{flutter_application_path}/.dart_tool/flutter_build/dart_plugin_registrant.dart'
+            flutter_filesystem = ''
+            flutter_app_dir = os.path.normpath(f'{source_dir}/{flutter_application_path}')
+            dart_plugin_registrant_file = f'{flutter_app_dir}/.dart_tool/flutter_build/dart_plugin_registrant.dart'
             if os.path.isfile(dart_plugin_registrant_file):
-                flutter_source_file = f'--source file://{dart_plugin_registrant_file}'
+                # The registrant is the one library with no package: URI, so it
+                # is named to the frontend by path -- and that path becomes the
+                # generated library's import URI, which the AOT image keeps
+                # verbatim. Naming it absolutely puts WORKDIR, and so TMPDIR,
+                # inside every shipped libapp.so and trips the buildpaths QA
+                # check. A relative path does not help: the frontend
+                # canonicalizes it back to an absolute file URI. Only a scheme
+                # with no absolute form does, so declare the app as a filesystem
+                # root and name the registrant against it. See #819.
+                registrant_scheme = 'org-dartlang-root'
+                registrant_uri = f'{registrant_scheme}:///' \
+                    f'{os.path.relpath(dart_plugin_registrant_file, flutter_app_dir)}'
+                flutter_filesystem = f'--filesystem-root {flutter_app_dir}' \
+                    f' --filesystem-scheme {registrant_scheme}'
+                flutter_source_file = f'--source {registrant_uri}'
                 flutter_source_package = '--source package:flutter/src/dart_plugin_registrant.dart'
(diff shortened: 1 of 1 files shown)


jwinarske:
**Verified on real Yocto artifacts.** Everything above was measured on emb-built images configured to match meta-flutter's flags; this closes that gap.

### The leak, in shipped output

`strings` on the `core-image-weston` rootfs of an existing master build (`ihs-v3-build`, TMPDIR `/mnt/raid10/yocto/master/ihs-v3-build/tmp`):

| app in rootfs | TMPDIR refs |
|---|---|
| `appstream-dart-example-flathub-catalog` | **1** — the registrant path |
| `flatpak-dart-example-flutter-remote-manager` | **1** — the registrant path |
| `packagekit-dart-example-packagekit-catalog` | 0 |
| `bluez-native-example-flutter-ble-scanner` | 0 |

One per plugin app, and it is the `dart_plugin_registrant.dart` absolute path. The two at 0 have no registrant — so nothing else leaks, which is the other half of the claim.

### Before/after on the real toolchain

Rather than rebuild (the tree is stale enough that bitbake wanted to redo glibc and the native toolchain first), I replayed the recipe's own recorded `do_compile` commands using the workdir's dart runtime, frontend server and gen_snapshot — the last still confirming meta-flutter's flag set is `--strip` with no `--obfuscate`.

The control replay reproduces the shipped artifact **byte for byte**:

```
shipped  : fe130e5afa69ed787bc54fa5f5fdb2f9…
replay A : fe130e5afa69ed787bc54fa5f5fdb2f9…   IDENTICAL
```

so the replay is the recipe, and the treated run is what the recipe will now produce:

| | TMPDIR refs |
|---|---|
| A — as shipped today | **1** |
| B — with the change | **0** |

```
A: file:///mnt/raid10/yocto/master/ihs-v3-build/tmp/work/…/dart_plugin_registrant.dart
B: org-dartlang-root:///.dart_tool/flutter_build/dart_plugin_registrant.dart
```

Registration is intact: `_PluginRegistrant` is present in both, under a library hash derived from the URI (`@437429399` → `@437393342`), and the unique-string counts are 21221 vs 21220 — the whole difference is the URI.

### Note on `buildpaths` in CI

The wrynose CI logs show no `QA Issue` lines before *or* after, so that workflow was never reporting this — `buildpaths` is a warning there, and the console log carries none. #566's report came from elsewhere. CI green means the build works; the evidence that the leak is fixed is the artifact comparison above.

PR #824, opened 2026-08-29
license: use "&" as the LICENSE operator, not "AND"

### LINKED ISSUE #823: LICENSE values use "AND"/"OR" instead of "&"/"|", aborting parse on current oe-core
### Summary

Nine recipes declare `LICENSE` with the SPDX prose operator `AND` (e.g.`LICENSE = "BSD-3-Clause AND MIT"`). BitBake `LICENSE` expressions only recognise `&`, `|` and `()` as operators, a bare `AND` is parsed as a licence literally named "AND". This trips the `license-format` QA check.

On older oe-core `license-format` was only a warning. Current oe-core lists it in `CHECKLAYER_REQUIRED_TESTS`, which `insane.bbclass` folds into the default `ERROR_QA`, so it is now fatal and **halts parsing of the whole layer**.

Image

AhmedAdelWafdy7:
This PR fixes #823.
Convert the `LICENSE` value of nine recipes from the SPDX prose operator `AND` to BitBake's `&` operator. One line per recipe; the licence set of each recipe is unchanged.

### DIFF
--- a/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-samples-ads_0.0.1.bb
+++ b/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-samples-ads_0.0.1.bb
@@ -11,7 +11,7 @@ HOMEPAGE = "None"
 BUGTRACKER = "None"
 SECTION = "graphics"
 
-LICENSE = "Apache-2.0 AND BSD-3-Clause AND OFL-1.1"
+LICENSE = "Apache-2.0 & BSD-3-Clause & OFL-1.1"
 LIC_FILES_CHKSUM = "file://LICENSE;md5=b7eeb61b41ae366e94383bca5e113fce"
 
 SRCREV = "ae636d23deae83fd0e7fec9b862a7fcdf2bcfdd8"
--- a/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-samples-crossword_0.1.0.bb
+++ b/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-samples-crossword_0.1.0.bb
@@ -11,7 +11,7 @@ HOMEPAGE = "None"
 BUGTRACKER = "None"
 SECTION = "graphics"
 
-LICENSE = "Apache-2.0 AND BSD-3-Clause AND OFL-1.1"
+LICENSE = "Apache-2.0 & BSD-3-Clause & OFL-1.1"
 LIC_FILES_CHKSUM = "file://LICENSE;md5=b7eeb61b41ae366e94383bca5e113fce"
 
 SRCREV = "ae636d23deae83fd0e7fec9b862a7fcdf2bcfdd8"
--- a/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-samples-multiplayer_0.0.1.bb
+++ b/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-samples-multiplayer_0.0.1.bb
@@ -11,7 +11,7 @@ HOMEPAGE = "None"
 BUGTRACKER = "None"
 SECTION = "graphics"
 
-LICENSE = "Apache-2.0 AND BSD-3-Clause AND OFL-1.1"
+LICENSE = "Apache-2.0 & BSD-3-Clause & OFL-1.1"
 LIC_FILES_CHKSUM = "file://LICENSE;md5=b7eeb61b41ae366e94383bca5e113fce"
 
 SRCREV = "ae636d23deae83fd0e7fec9b862a7fcdf2bcfdd8"
--- a/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-templates-basic_0.0.1.bb
+++ b/meta-flutter-apps/recipes-graphics/flutter-apps/first-party/flutter-games-templates-basic_0.0.1.bb
@@ -11,7 +11,7 @@ HOMEPAGE = "None"
 BUGTRACKER = "None"
 SECTION = "graphics"
 
-LICENSE = "Apache-2.0 AND BSD-3-Clause AND OFL-1.1"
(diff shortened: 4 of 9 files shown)


jwinarske:
Thanks for the patch — the conversion is clean and complete (all nine recipes that use `AND`, one line each, licence sets unchanged). But I would like to **hold this**, because on the oe-core master builds against, it moves the recipes the wrong way. Details and the timeline are in #823.

Short version: oe-core [`51c7930220`](https://git.openembedded.org/openembedded-core/commit/?id=51c7930220) (2026-07-15) added the SPDX licence library. Since then `AND` is the native operator and `&` is legacy — `_substitute_legacy_license()` rewrites `&` → `AND`, and `license.py:498` warns *"is using an old syntax and should be upgraded to"* exactly when that rewrite fires. Verified against oe-core master `02e3265178`:

| expression | "old syntax" warning |
|---|---|
| `Apache-2.0 AND MIT` | no |
| `Apache-2.0 & MIT` | **yes** |

So merging this would take nine currently-clean recipes and make each emit a warning. The `AND` was introduced deliberately in `659b662b` for that reason.

The behaviour you hit is real on **older** oe-core, where the operators were only `&|()` and a bare `AND` parsed as a licence name. That is why every release branch (wrynose, scarthgap, kirkstone, dunfell) uses `&` and has zero recipes with `AND` — they pair with older oe-core, and there `AND` genuinely fails.

Before closing this I would like to know which oe-core revision and meta-flutter branch you built — asked in #823. If it turns out you were on tip-of-master oe-core and still saw the failure, this PR may be right and my analysis wrong, so I would rather confirm than assume.

Independently of the outcome: the layer nowhere records that master intentionally differs from the release branches on this operator, which is what made the report look like an obvious bug. That is a genuine gap and I will document it either way — so the report was useful even if the patch does not land.

jwinarske:
One more thing on this, and it applies **whichever way #823 resolves** — so it is worth knowing even if you turn out to be right about the operator.

Six of the nine recipes here are **generated**, from `meta-flutter-apps/conf/flutter-apps.json`:

```json
{
    "uri": "https://github.com/flutter/games.git",
    "license_type": "Apache-2.0 & BSD-3-Clause & OFL-1.1",
}
```

That single entry produces all six `flutter-games-*` recipes. So patching the `.bb` files does not hold — the next `roll_meta_flutter.py` run regenerates them from the manifest and overwrites the change. To move those six you have to move the manifest entry too.

Note which way that entry points: it still says `&`. The recipes say `AND` only because `659b662b` hand-edited the generated output and never updated the manifest — my oversight, not yours. So master already has a latent regression here, independent of this PR, and your patch would have been silently reverted on the next roll.

There is also no check to catch the drift: `roll_meta_flutter.py:129-141` validates the licence *set* against the licence text, which is a real check, but never looks at the operator. `Apache-2.0 AND MIT` and `Apache-2.0 & MIT` both pass.

I have opened #831 for the generator side — manifest, operator validation, and a dead `detect_license()` helper that hard-codes `&` with a now-incorrect docstring. That work is worth doing regardless of the outcome here.

So the state of this PR: the three hand-written recipes (`libwebrtc`, `rive-text`, `ivi-homescreen-keyboard-test`) are the only ones a recipe-level patch can actually fix, and whether they should change at all depends on the oe-core question in #823. Still waiting on your revision/branch there before deciding.

jwinarske:
**Correction to my previous comment.** I said `flutter-apps.json` still declares `Apache-2.0 & BSD-3-Clause & OFL-1.1` for flutter/games. That is wrong — I read the manifest out of a working tree checked out on a scarthgap branch and reported it as master's. Master's manifest reads:

```json
"uri": "https://github.com/flutter/games.git",
"license_type": "Apache-2.0 AND BSD-3-Clause AND OFL-1.1",
```

So master is internally consistent: manifest and generated recipes both say `AND`, and there is no pre-existing drift. Apologies for the noise.

The conclusion for this PR is unchanged, but the reason inverts. Six of the nine recipes are generated from that manifest entry, and the entry says `AND` — so a patch that changes only the `.bb` files to `&` is overwritten the next time `roll_meta_flutter.py` runs. To move those six you would need to change the manifest as well.

The three hand-written recipes (`libwebrtc`, `rive-text`, `ivi-homescreen-keyboard-test`) are the only ones a recipe-level patch holds on its own.

#831 has been corrected accordingly — what remains there is that the tool never validates the operator at all, which is a real gap regardless of which spelling is right.

Still the open question, from #823: which oe-core revision and which meta-flutter branch did you build? That decides whether any of this should change.

jwinarske:
New evidence, from CI rather than from reading the parser.

#839 removed the layer-wide `license-format` demotion in `ci-build.inc` — the check is now fatal on master CI, as it is by default in oe-core. That run is **green** with the nine `AND` recipes exactly as they are:

https://github.com/meta-flutter/meta-flutter/actions/runs/33293516802

Before #839 the check was demoted to a warning, so a green run proved nothing either way — which is precisely why I could not answer your report from CI results and had to work it out from `license.py` by hand. Now the check that would flag a bad LICENSE expression is enabled, and it does not flag these.

So on oe-core master, `AND` is accepted and `&` is the form that draws the "old syntax" warning. That is the opposite of what this PR does, and I do not think it should land against master as written.

The behaviour you reported is still real on older oe-core, and every release branch here uses `&` for that reason. If you were building meta-flutter master against a release-branch or pre-2026-07-15 oe-core, that combination is unsupported and would produce exactly what you saw — worth confirming, and if you were on tip-of-master and still hit it I would genuinely like the revision, because then something is wrong with my analysis rather than with the report.

Either way the report was useful: it exposed that nothing documented the split (now #835), that the roll tool never checked the operator (now #836), and that CI had this check switched off (now #839). Thanks for pushing on it.

jwinarske:
Not needed. #823 turned out to be meta-flutter master built against oe-core
yocto-6.0.2 (wrynose).

master pairs with oe-core master, where `AND` is the SPDX operator and `&` is
legacy — 659b662b converted these nine recipes to `AND` to clear the "old
syntax" warnings. This reverses that.

Closing unmerged.

Issue #675, opened 2025-06-24
Build with `--no-pub`


edrose:
We've had some odd issues that very occasionally come up when we are building our flutter app on 3.32.2 (but I think it's also valid for other versions). It only appears for some combinations of `pubspec.yaml`, `pubspec.lock`, and the state of `pub.dev` so it feels pretty random and difficult to track down.

1. During the build process `flutter pub get --enforce-lockfile` is run to get the pub cache, which is then archived
2. The archive is unpacked and `flutter pub get --enforce-lockfile --offline` is run
3. `flutter build bundle` is then run during the compile step. Flutter tries to access the network and fails, causing the whole build to fail

It turns out that this occurs because, by default, flutter will do a `pub get` before a build. Since there is no way to pass the `--offline` or `--enforce-lockfile` flags to `flutter build`, it tries to access the network and fails to do so.

The solution we have to this is to override `FLUTTER_BUILD_ARGS` to equal `bundle --no-pub` which prevents flutter from doing a `pub get` before the build.

I wanted to open a PR for this but there is a catch. The `--no-pub` option is not for `flutter build`, it's defined on the subcommands individually so `flutter build --no-pub bundle` isn't valid, it has to be `flutter build bundle --no-pub`. I wasn't sure how often the `FLUTTER_BUILD_ARGS` variable was overridden so it didn't feel like it could be hard-coded.

It feels like an essential option for building reliably in yocto so it feels like it _should_ be hard-coded. I just wasn't sure the best way to achieve that.

jwinarske:
Hi @edrose

Using `--offline` is "supposed" to only use the local pub repository (which was created in the archive step), and "should not" attempt to use the the network.  If you are seeing it is, and that `--no-pub` is reliable in all cases (lots of tricky corner cases) then a change makes sense.  I think until we can prove out all the corner cases, you can simply do this:

FLUTTER_BUILD_ARGS is defined as
```
FLUTTER_BUILD_ARGS ??= "bundle"
```

Which allows you to override it in your app recipe
```
FLUTTER_BUILD_ARGS = "bundle --no-pub"
```

jwinarske:
This is implemented on every branch, and in a way that avoids the hard-coding problem you raised.

Rather than putting `--no-pub` into `FLUTTER_BUILD_ARGS` — where a recipe overriding that variable would silently drop it — `common.inc` appends it based on whether the compile task is allowed network at all:

```python
[...]

PR #842, opened 2026-08-30
tools: add pubvendor, generating a pub cache SRC_URI fragment from pubspec.lock

jwinarske:
Phase 1: land the generator and its consuming class, standalone and callable by hand. Wiring it into `roll_meta_flutter.py` is deliberately left for a follow-up.

## What it is for

An app's pub dependencies are fetched by `pub get` during the build. That means a Flutter app recipe cannot honour `BB_NO_NETWORK`, and none of those inputs are checksummed by bitbake. `pubvendor.py` turns `pubspec.lock` into an `.inc` of `SRC_URI` entries, so they arrive through do_fetch/do_unpack like any other source, with the lockfile's own sha256 values as the checksums.

```
python3 tools/pubvendor/pubvendor.py -i app/pubspec.lock -o recipes-apps/app/app-pubcache.inc
```

```
require app-pubcache.inc
inherit pub-cache
PUBSPEC_APP_DIR = "${S}"
```

## Design points worth reviewing

- **No network on the default path.** Hosted archive URLs are built from the pub API layout and checksums come from the lockfile, so generation is deterministic and offline. Only a pre-Dart-2.19 lockfile needs `--resolve-missing`, which fetches once at generation time and pins the digest into the output rather than deferring it to build time.
- **Git deps pin `resolved-ref`, not `ref`.** A moving branch cannot change what a recipe builds. One bare clone serves every package from the same repo.
- **The class fails closed.** It refuses to build if the in-tree `pubspec.lock` no longer matches the one the `.inc` was generated from, so a stale fragment is an error rather than a silent mismatch.
- **The staged layout is a pub internal**, not a documented contract. It is asserted by the tests and is meant to be validated per-SDK by an offline `pub get` job rather than assumed stable — worth deciding whether that job belongs in this layer's CI.
- `--style old` emits `_append` for dunfell.

Clean-room: the pub2yocto project supplied the concept, none of its code.

## Verification

`14 passed` on the included suite. Beyond that, run against a real 46-package lockfile (`layer_playground`):

```
42 hosted, 0 git, 4 skipped
# Skipped (not fetched): flutter (sdk), flutter_test (sdk), flutter_web_plugins (sdk), sky_engine (sdk)
```

Emitted checksums spot-checked against the lockfile (`async`, `clock`, `characters` — all match), and the output is byte-identical across runs.

## Placement and SPDX

- `tools/pubvendor/` for the generator and tests, matching the layout its tests expect
- `classes/pub-cache.bbclass` — the regular bbclass location, so `inherit pub-cache` resolves
- Headers brought in line with the layer: `SPDX-FileCopyrightText` + `SPDX-License-Identifier` for the Python, and the `Copyright (c)` + identifier shape `gn-fetcher.bbclass` uses for the class

## Follow-ups, not in this PR

1. **Call it from the roll** so app recipes get their fragment generated. The seam is `create_recipes.py`, which is where `roll_meta_flutter.py` already writes recipes.
2. **Run the tests in CI.** `tools/**` is in the build workflow's `paths-ignore`, so nothing runs them today; they need a separate lightweight job.
3. **An offline `pub get` matrix** to validate the cache layout per Dart SDK, as the script's own notes call for.

Relevant to #675, which is about the same underlying problem from the other direction.


jwinarske:
Follow-ups filed, in dependency order:

- **#844** — nothing runs `tools/pubvendor/tests/`. `tools/**` is in the build workflows' `paths-ignore` (correctly — a Python change should not trigger four Yocto builds), so the 14 tests added here have no runner. Needs a separate lightweight job.
- **#845** — validate the staged cache layout with an offline `pub get`, per Dart SDK. The layout is a pub internal, not a contract; the unit tests can only assert what we believe it to be, and only a real `pub get --offline` proves the belief still holds.
- **#843** — call the generator from the roll. Blocked on #844: the roll should not depend on a generator that nothing tests.

jwinarske:
Additional verification: the generated fragment produces a cache a real Flutter app can `pub get` from, offline.

Staged a pub cache containing **only** what the fragment names — fetched each archive, verified its checksum against the `.inc`, unpacked into the `subdir` layout, synthesized the `hosted-hashes` files the class writes — then ran exactly what `do_pub_get_offline` runs:

```
$ PUB_CACHE=<staged only> dart pub get --offline --enforce-lockfile
Resolving dependencies...
Got dependencies!
```

**40 of 40 checksums matched** the digests pubvendor took from the lockfile. That is the concrete evidence for the no-network claim: the digests are correct without the generator ever fetching.

Resolution split from the resulting `package_config.json`:

| source | count |
|---|---|
| staged cache | 40 |
| Flutter SDK (`flutter`, `flutter_test`, `flutter_web_plugins`, `sky_engine`) | 4 |
| the app itself | 1 |

The 4 are the `source: sdk` entries the generator skips, and skipping them is right — they come from the Flutter SDK tree, which a Yocto build already stages via `flutter-sdk-native`. Worth noting that plain `dart pub get` resolves them; `flutter pub get` is not needed, which I had expected might be a problem and is not.

**Still unproven, and I would not want this read as more than it is:** that app has no git dependencies (40 hosted, 0 git). The git path — `resolved-ref` pinning, the `<repo>-sha1(url)` cache directory, scp-style normalization — has unit tests but no offline `pub get` behind it, and it is the more fragile half since git cache keying is the least documented part of the contract. One Dart SDK only, 3.13.1.

Recorded as a baseline on #845 so the job there has a known-good result to automate against, with those two gaps named as its minimum scope.

Issue #843, opened 2026-08-30
roll: generate the pub cache fragment for app recipes


jwinarske:
Phase 2 of #842. The generator landed standalone; this is wiring it into the roll so app recipes get their fragment without anyone remembering to run it.

## Where it goes

`roll_meta_flutter.py` already resolves each app and hands off to `create_yocto_recipes()` in `create_recipes.py`, which is what writes the recipe. That is the seam: after the app's source is checked out and before/alongside the recipe write, run `pubvendor.parse_lockfile()` on the app's `pubspec.lock` and emit the `.inc` next to the generated recipe, then have the recipe `require` it and `inherit pub-cache`.

## Decisions to make first

- **Which apps.** Every generated app recipe, or opt-in per entry in `flutter-apps.json`? A fragment is only useful to a recipe that also inherits `pub-cache`, and turning that on for ~25 apps at once changes how all of them fetch.
- **Lockfile availability.** The roll checks out app sources, so `pubspec.lock` is present — but an app without a committed lockfile has nothing to vendor, and that needs to be a clear skip rather than a failure.
- **`--style` per branch.** dunfell needs `_append`; kirkstone and newer need `:append`. The roll knows which branch it is on only implicitly. Same branch-specific-syntax problem the LICENSE operator had in #831/#836, and worth solving the same way — one declared constant, checked.
- **Regeneration churn.** The `.inc` changes whenever a lockfile does, so a roll will produce large diffs. Acceptable, but worth knowing before it lands.

## Suggested sequencing

Wire it for a single app first, prove an offline build of that recipe, then widen. Turning it on layer-wide before an offline build has actually been demonstrated would make a broad change on an unproven assumption.

Depends on #844 (CI coverage for the tool) being in place first, so a change to the generator cannot silently break every app recipe.

Related: #675, #842.

jwinarske:
Started this and stopped before writing code, because the layer already has a pub cache mechanism and pubvendor collides with it. Wiring the generator in as this issue originally described would produce a recipe that fights itself. Recording what is in the way.

## What exists today

`common.inc` populates the pub cache at build time:

- `do_archive_pub_cache` (`addtask ... after do_patch before restore_pub_cache`) runs `flutter pub get --enforce-lockfile` **with network**, then archives `${WORKDIR}/pub_cache`
- `do_restore_pub_cache` (`before do_configure`) restores from a workspace-level `PUB_CACHE_ARCHIVE` when one exists

[...]

PR #852, opened 2026-08-30
roll: generate a pub cache fragment for apps that opt in

jwinarske:
Step 3 of #843. **Stacked on #851** — merge that first.

Opt-in and inert: no app sets `"pubvendor"` yet, so nothing changes for any existing recipe.

## Why it re-resolves the lockfile

For an opted-in app the roll deletes `pubspec.lock` and re-runs `flutter pub get` before generating the fragment. That looks destructive, and is the opposite:

The lockfile an app ships was resolved against whatever SDK its authors used, rarely the one this layer pins. That mismatch is exactly why every generated recipe has always carried `PUBSPEC_IGNORE_LOCKFILE = "1"`, which **deletes the lockfile at build time** so each build re-resolves. Two people building the same SRCREV a week apart can get different dependency sets today.

Resolving once at roll time and committing the fragment is that same concession made **once, visibly, and identically for everyone** — strictly more reproducible than the status quo, not less. Verified: the re-resolved lockfile records `dart: ">=3.13.1 <4.0.0"`, the pinned SDK.

An opted-in recipe therefore sets `PUBSPEC_IGNORE_LOCKFILE = "0"`; the fragment is generated from that lockfile and `pub-cache.bbclass` refuses to build if the in-tree one no longer matches, so it has to survive `do_unpack`.

## Two failures handled rather than propagated

- **No flutter on PATH** — reported and skipped. The SDK is only needed by apps that opt in, so a roll with none must not require it.
- **`pub get` cannot resolve against the pinned SDK** — a fact about that app, and one of them must not take down a roll of twenty-five. Warns and moves on.

In both cases the recipe is written **without** the `require` line. A recipe requiring a fragment that was never generated is a parse error for the entire layer — much worse than one app missing its fragment. That is why generation happens before the recipe is written and its result gates the emission.

## Branch-specific style

`OVERRIDE_STYLE` is new in `common.py`, declared next to `LICENSE_OPERATOR` because it is the same kind of fact: branch specific, correct on its own branch, a regression if carried to the other. dunfell will set it to `old`.

## Verified

Ran the generation path against a real 40-package app:

```
demo-app-pubcache.inc: 40 hosted, 0 git, 4 skipped
:append entries present: yes    _append entries: 0
```

and the no-flutter path:

```
** ERROR: x: "pubvendor" needs flutter on PATH to resolve against the pinned SDK **
result: False
```

Tools suite: 14 passed.

## Still outstanding in #843

Opting one app in and proving it with a real `BB_NO_NETWORK` build. Everything to this point is reasoning about task graphs and generated text; only a build shows whether an app with plugins comes out with its plugins.

### DIFF
--- a/classes/pub-cache.bbclass
+++ b/classes/pub-cache.bbclass
@@ -71,9 +71,27 @@ python do_write_hosted_hashes() {
 }
 addtask write_hosted_hashes after do_unpack before do_configure
 
+# `flutter pub get`, not `dart pub get`. Both resolve offline from the staged
+# cache, but only the flutter one writes .flutter-plugins-dependencies, which
+# is what tells the build which plugins to register and feeds the generated
+# dart_plugin_registrant.dart. With `dart pub get` an app builds green and
+# registers no plugins.
 do_pub_get_offline() {
     cd ${PUBSPEC_APP_DIR}
-    dart pub get --offline --enforce-lockfile
+    flutter pub get --offline --enforce-lockfile
 }
 addtask pub_get_offline after do_check_pubspec_lock do_write_hosted_hashes before do_compile
 do_pub_get_offline[network] = "0"
+
+# The layer's own pub cache path fetches with the network and carries the
+# resolution artifacts through an archive (common.inc, the .project copy in
+# do_archive_pub_cache and the moves back in do_restore_pub_cache). A recipe
+# staging its cache through SRC_URI needs neither: do_unpack puts the packages
+# in place and do_pub_get_offline regenerates the artifacts locally. Leaving
+# both enabled would mean two populators of ${WORKDIR}/pub_cache, one of them
+# reaching for the network, which is the thing this class exists to avoid.
+#
+# Doing it here rather than in common.inc keeps opting in to exactly
+# "inherit pub-cache", with no effect on any recipe that does not.
+do_archive_pub_cache[noexec] = "1"
+do_restore_pub_cache[noexec] = "1"
--- a/classes/pub-cache.bbclass
+++ b/classes/pub-cache.bbclass
@@ -22,6 +22,27 @@ PUBSPEC_APP_DIR ?= "${S}"
 
 export PUB_CACHE
 
+# The fragment is generated from a lockfile resolved against the SDK this
+# layer pins, which is not the one the app committed upstream. do_unpack
+# brings that resolved lockfile in through SRC_URI; install it over the app's
+# own before anything reads it, so what gets built is what was vendored.
+python do_install_pubspec_lock() {
+    import os
+    import shutil
+
+    name = d.getVar('PUBSPEC_LOCK_FILE')
+    if not name:
+        return
+    src = os.path.join(d.getVar('WORKDIR'), name)
+    if not os.path.isfile(src):
+        bb.fatal("PUBSPEC_LOCK_FILE %s not found in WORKDIR; is it in SRC_URI?"
+                 % name)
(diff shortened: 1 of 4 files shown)


jwinarske:
Pushed a fix for a hole I found while setting up the local build. Worth describing, because the build would have failed on it immediately and the cause is not obvious from the code.

The fragment records `PUBSPEC_LOCK_SHA256` of the lockfile it was generated from, and `pub-cache.bbclass` refuses to build when the app's lockfile does not match. **Those could never match.** The roll re-resolves in its own clone; the recipe fetches the app at SRCREV and gets whatever lockfile upstream committed:

```
original, from git SRCREV : 325a385376c5329a…
regenerated, pinned SDK   : 84edb5e81f591f8a…
recorded in the .inc      : 84edb5e81f591f8a…
```

Every opted-in recipe would have stopped at `do_check_pubspec_lock`.

The fix: write the resolved lockfile beside the fragment, carry it into the build through `SRC_URI`, and install it over the app's own in a new task ordered before the check. What gets built is then what was vendored, instead of two descriptions of the app disagreeing.

The check keeps its purpose rather than being defeated — it now catches the layer's own `.inc` and `.lock` drifting apart, which is exactly the case a regeneration fixes.

Verified the two now agree on the hash, so the comparison passes rather than aborting.

jwinarske:
Ran the local build. It found two more defects and got most of the way; recording where it stopped and why.

## Two more fixes, pushed

**`file://` unpacks to `UNPACKDIR`, not `WORKDIR`** — from scarthgap on. The class looked only in `WORKDIR`, so the vendored lockfile was never found and `do_install_pubspec_lock` stopped on its own error. Now takes whichever the release provides, since dunfell and kirkstone have no `UNPACKDIR`.

**`do_pub_get_offline` ran with no `flutter` on PATH.** The SDK is staged by `flutter-sdk-native`, but nothing puts it on a task's PATH — `common.inc` builds an env for its own pub invocations and this class was not getting the benefit. Now sets the same three things it does: SDK and pub cache `bin` on PATH, `HOME` and `XDG_CONFIG_HOME` at `WORKDIR`.

## What the build did prove

- The opted-in recipe **parses**, and costs 43 tasks rather than a rebuild of the world
- **`bitbake --runall=fetch` pulled 156 pub tarballs into `DL_DIR`** through `SRC_URI`, with bitbake verifying every checksum. That is the fragment doing its job
- With `BB_NO_NETWORK=1`, `do_install_pubspec_lock` and `do_check_pubspec_lock` both **pass** — so shipping the resolved lockfile alongside the fragment works

## Where it stopped, and why I am not claiming it as a defect

`do_pub_get_offline` then failed because `bin/flutter` was missing from that recipe's `recipe-sysroot-native`. That is almost certainly **my environment, not the change**: to avoid a multi-hour cascade I pinned oe-core, bitbake and meta-yocto back to pre-build revisions, which meant re-staging a sysroot under oe-core Aug-17 with meta-flutter Aug-30. Sibling recipes staged before that surgery all have `bin/flutter`; only the one I rebuilt under the mixed configuration does not.

So the pinning trick that made the build affordable also made the last step untrustworthy. Worth knowing for anyone repeating this: it is fine for measuring cost, not for concluding anything about staging.

## Still outstanding

The `pub get` step itself is unproven, and with it the plugin-registration assertion — which is the whole reason to run this. That wants either a tree consistent with current master, or the CI job from the plan. Given four defects surfaced here that neither unit tests nor review caught, I would not merge this series on inspection alone.

PR #855, opened 2026-08-30
appstream-dart-flathub-catalog: vendor the pub cache, and prove it in CI

### LINKED ISSUE #843: roll: generate the pub cache fragment for app recipes

jwinarske:
Closes #843. First app on the pubvendor path, and **the first time any of it is exercised by a build**.

Its dependencies now arrive through `SRC_URI` with checksums instead of a networked `pub get` inside `do_archive_pub_cache`.

## The proof

```
bitbake --runall=fetch "$APP"      # network: populate DL_DIR from SRC_URI
BB_NO_NETWORK=1 bitbake "$APP"     # anything undeclared fails here
```

Anything the fragment failed to declare cannot be fetched in the second pass, so it fails rather than working quietly on a machine that happens to be online.

**Offline is only half of it.** `dart pub get` also resolves offline but writes no `.flutter-plugins-dependencies`, so an app can build green having registered nothing — the defect fixed in #851, which no unit test could see. So the step also asserts the registrant is in the packaged image:

```
strings -a "$LIBAPP" | grep -q '_PluginRegistrant'
```

Both halves pinned: builds offline, *and* comes out with its plugins.

I chose this app because it registers three Linux plugins — `jni`, `path_provider_linux`, `url_launcher_linux` — so the assertion has something to catch. A plugin-free app would pass while proving nothing. Verified the `find` pattern and the `grep` against a real packaged `libapp.so` before committing.

## A sixth defect, found reading the recipe

`do_inject_user_defines` appends a `hooks:` section to `pubspec.yaml`, ordered `before do_archive_pub_cache` — which `pub-cache.bbclass` makes **noexec**. The class's own tasks were ordered only against `do_unpack` and `do_configure`, so nothing stopped the offline resolve running *before* that edit landed.

The recipe's own comment explains why that matters:

> Editing pubspec.yaml afterwards makes the offline pub get in do_compile re-resolve, which tries to fetch security advisories and fails with no network.

All three class tasks now run `after do_patch`.


jwinarske:
Rebased onto #856, so one four-machine round proves both.

The failure on the previous run was **not this change**. `bitbake --runall=fetch` succeeded — 578 tasks, the app's `do_fetch: Succeeded`, all 78 fragment entries fetched and checksummed in clean CI, which is the first time the fragment has worked in a real environment. The `BB_NO_NETWORK=1` step then failed at *parse*, in an unrelated layer:

```
ERROR: ExpansionError during parsing .../meta-vulkan/.../maplibre-vk_git.bb
Failure expanding variable fetcher_hashes_dummyfunc[vardepvalue],
expression was ${@bb.fetch.get_hashvalue(d)}
ERROR: Parsing halted due to errors
```

Four meta-vulkan recipes named a **tag** in `SRCREV`. bitbake cannot know what a tag points at without asking the remote, so it resolves them while parsing — making the parse require network, and halting it for the whole build when there is none. Fixed upstream in jwinarske/meta-vulkan#29 (tags pinned to shas), and #856 then drops the layer from CI entirely, since nothing master builds uses it.

So this round tests two things at once: that dropping meta-drm and meta-vulkan is safe, and that the vendored pub cache builds offline with its plugins registered.

Worth flagging for review regardless of the outcome: `BB_NO_NETWORK` is a blunt instrument. It applies to the whole parse, so *any* recipe in *any* layer that cannot resolve offline breaks it — meta-vulkan today, something else later. #856 removes today's exposure rather than the general fragility.

jwinarske:
CI got further and found a real limitation. **meta-vulkan is no longer in the way** — with #856 stacked, the parse error is gone and the build reached `do_pub_get_offline`, which is the furthest this has ever got.

It then failed:

```
| Got socket error trying to find package test at https://pub.dev.
| Failed to update packages.
ERROR: ... do_pub_get_offline: Execution of '...' failed
```

## What is established

- `test` is **not in the app's `pubspec.lock`**, so pubvendor never generated a `SRC_URI` entry for it — the fragment has 78 entries and `test` is not among them
- `test` **is** a `dev_dependency` of `appstream_dart`, the package this app depends on **by path** (`../..`), and which carries the native-assets `hooks:` section
- The same `flutter pub get --offline --enforce-lockfile` **succeeds on my machine**, because the ambient pub cache already contains `test`. It fails in CI because the staged cache is *only* what the fragment names

So the fragment is incomplete for this app: the build needs a package the lockfile does not name.

## What is not yet established

Exactly why pub wants it. The plausible reading is that resolving a path dependency whose package has a build hook pulls in that package's own dev dependencies, which the app's lockfile does not record. I have not proven that, and would not want it written down as fact.

That distinction matters, because it decides the fix. If the lockfile genuinely cannot describe an offline build for path-dependency-plus-hook apps, `pubvendor` needs another input; if pub is merely being asked the wrong question, the fix is in how the class invokes it.

## The process failure worth naming

`offline_check.py` exists precisely to catch this — it stages only what the fragment names and fails if anything resolves elsewhere. I ran it against the synthetic fixture and never against the app I actually opted in. Had I pointed it at this app, it would have failed in about twenty seconds instead of costing a four-machine round.

That is the seventh defect in this series, and the second where the tooling to catch it already existed and was aimed at the wrong target.

## Suggested next step

Do not merge this. Point `offline_check.py` at the real app, reproduce in seconds, and settle the mechanism before choosing a fix. #856 is independent and good on its own evidence — it should be unstacked and merged separately.

jwinarske:
Mechanism found, and it is not what the error suggests.

**The fragment is complete.** Staging exactly the 78 packages it names into an empty cache and running CI's exact command succeeds:

```
$ PUB_CACHE=<only the fragment's 78> flutter pub get --offline --enforce-lockfile
Resolving dependencies...
Got dependencies!
```

**`test` is flutter's, not the app's.** `packages/flutter_tools/pubspec.yaml:124` declares `test: 1.31.1`. On a fresh `HOME` and `PUB_CACHE`, `flutter pub get` resolves the tool's own dependencies before it gets to the app's — that is what reached pub.dev, and why the missing package is one the app never mentions.

**The layer already solves this**, and the class does not. `common.inc:170`:

```
# clean existing pub-cache, populate from the flutter-sdk-native pub-cache
```

The existing path seeds `${WORKDIR}/pub_cache` from the SDK's own cache before running any flutter command. `pub-cache.bbclass` points `PUB_CACHE` at a tree holding only the app's dependencies, so the tool's are absent.

So the vendored cache has to **add to** the SDK's tool cache rather than replace it. That is a design correction, not a patch: the fragment describes the *app's* dependencies, and the tool's are a separate concern the layer already knows how to supply.

## Why this cost a CI round it should not have

`offline_check.py` stages into an empty cache and would never have caught this, because it runs `dart pub get` outside a Yocto build with an SDK whose tool cache is already warm. The check I built to prove offline resolution cannot see the one thing that differs in the real environment.

That is worth recording as a limitation of the check itself, not just of this app.

## Where that leaves the PR

Still do not merge. The fix belongs in `pub-cache.bbclass`, seeding from the SDK cache the way `common.inc` does, and the assertion in `offline_check.py` needs rethinking if it is to be evidence for anything beyond its own fixture.

#856 is unaffected and good to merge on its own.

jwinarske:
Rebased onto master now that #857 has landed. The class changes that were in this branch are gone — #857 carries them — so this is now just the app opt-in, its fragment and lockfile, and the CI step.

What #857 fixed is exactly what failed here: the staged cache held only the app's dependencies, so `flutter pub get` could not resolve `flutter_tools`' own — including `test`, the package the error named and which appears nowhere in the app's lockfile. The class now merges the SDK's `.pub-cache` in, staged copies winning.

This run is therefore the same test as before with that cause removed. It still proves both halves: the app builds with the network off, and the packaged `libapp.so` carries `_PluginRegistrant`.

jwinarske:
Diagnosed, and it is architectural rather than another one-line fix. I would stop patching and reconsider the approach.

## What this run showed

The analytics fix worked — flutter's banners are gone. `do_seed_pub_cache` ran and succeeded. The failure is now:

```
ClientException with SocketException: Failed host lookup: 'pub.dev'
uri=https://pub.dev/api/packages/archive/advisories
```

`archive` is not in the app's lockfile or fragment. It is a **flutter_tools dependency**, like `test` before it. Both are present in the seeded cache (`archive-3.6.1`, `test-1.31.1` are in `flutter-sdk-native`'s `.pub-cache`), so this is not a missing package — pub is fetching their **security advisories**, which is a separate request and one `--offline` does not suppress.

## Why pub is resolving at all

Advisories are fetched when pub resolves. The app's own resolution is satisfied from the lockfile — proven locally, with the network unshared away entirely:

```
$ unshare -rn -- flutter --suppress-analytics pub get --offline --enforce-lockfile
Got dependencies!
```

That succeeds against only the 78 staged packages. So it is not the app's resolve. It is **flutter's own tool bootstrap**: the first `flutter` invocation in a fresh environment re-resolves `flutter_tools`, and that resolution triggers the advisory fetch for its dependencies.

## The layer already avoids this, and this class removed the thing that did

`do_archive_pub_cache` runs `flutter config`, `flutter clean` and `flutter pub get` **with network**. That is where the tool bootstrap happens, once, online — after which offline invocations reuse it. This class marks that task `noexec` and then runs flutter for the first time in a task that has no network.

So the offline app build needs the SDK to arrive **already bootstrapped**, and today it does not: the layer papers over that by doing a networked `pub get` per app.

## Options, none of them a patch to this PR

1. **Bootstrap the tool in `flutter-sdk-native`**, so the staged SDK needs no online first run. The real fix, and it benefits every consumer, not just vendored-cache recipes. Matches the instinct behind doing the analytics opt-out where there is a network.
2. **Keep a networked bootstrap task in the class**, replacing what `do_archive_pub_cache` did without its cache wipe. Works, but re-introduces a network dependency into a path whose purpose is not having one.
3. **Find whether pub can be told to skip advisories.** If such a flag exists this becomes small; I have not found one.

I would not merge this PR in its current form. The vendored cache works — that much is now proven, locally and by `--runall=fetch` in CI. What does not work is running flutter offline against an SDK that has never been run online, and that is a property of the SDK recipe rather than of this app or this class.

jwinarske:
Correction to my previous comment, and the actual root cause. It is **not** the flutter tool bootstrap — I was wrong about that.

`flutter-sdk-native` already bootstraps properly: `flutter update-packages`, `flutter doctor`, the config calls and three `flutter create` templates, all with `PUB_CACHE` pointed at the SDK's `.pub-cache`. And that state survives staging intact:

```
flutter_tools.stamp : 6655482ec06e547f90abf8ae7590466f4415978d
staged git HEAD     : 6655482ec06e547f90abf8ae7590466f4415978d
pubspec.yaml newer than pubspec.lock : no
.git, flutter_tools.snapshot, packages/flutter_tools/.dart_tool : all present
.pub-cache : 583M
```

So the tool is genuinely fresh and does not rebuild. Nothing to fix there.

## What is actually happening

`appstream_dart` — the package this app depends on **by path** — carries a native-assets `hook/build.dart`, and has **no lockfile of its own**. Running that hook requires resolving *its* dependencies, and that is a fresh resolution: it fetches security advisories, which needs the network.

That resolution pulls in things the app's lockfile never mentions:

| package | where it comes from |
|---|---|
| `test` | `appstream_dart` dev_dependency |
| `archive` | transitively, from that same graph |

Both are in the SDK's `.pub-cache` already, which is why seeding changed the symptom from `test` to `archive` without fixing anything — the packages were never the problem, the *resolution* is.

## The limitation this exposes

**An app's `pubspec.lock` is not a complete description of what its offline build needs when a path dependency carries a build hook.** The hook's package has its own dependency graph, resolved separately, and pubvendor's one-lockfile-per-app model does not cover it.

That is a real constraint on the approach, not a bug in this class.

## What is still proven

The vendored cache works. For the app's own resolution, offline, against a cold cache with only the 78 staged packages:

```
$ unshare -rn -- flutter --suppress-analytics pub get --offline --enforce-lockfile
Got dependencies!
```

and `--runall=fetch` fetched all 156 tarballs in clean CI with checksums verified.

## Recommendation

This app was the worst possible first choice — I picked it for having three registered plugins, without noticing it also has a path dependency with a hook and no lockfile. Two ways forward:

1. **Opt in a simpler app first** — plugins, but no path-dependency-with-hook — and land the mechanism on something it actually covers.
2. **Extend pubvendor to vendor a hook package's resolution too**, which means generating a lockfile for it. Larger, and worth its own issue.

I would do 1 now and file 2. This PR should be closed or retargeted at a different app.

jwinarske:
Two corrections, then a diagnostic instead of another theory.

## The SDK pub cache does restore correctly

I suggested it might not. It does:

| recipe sysroot | files in .pub-cache | staged |
|---|---|---|
| flatpak-dart | 22,978 | 2026-08-19 |
| bluez-native | 22,978 | 2026-08-19 |
| packagekit-dart | 22,978 | 2026-08-19 |
| appstream-dart | **0** | 2026-08-30 |

The empty one is the sysroot my own layer-pinning experiment re-staged today — the same corruption that made `bin/flutter` disappear from it earlier. Every sysroot from a normal build is complete. So `do_seed_pub_cache` had real content to seed from, and that is not the cause.

## Everything I can reproduce locally passes

With CI's staged flutter binary, CI's merged cache (78 vendored + the SDK's, 636M), every `.dart_tool` cleared, a fresh HOME, and the network unshared away entirely:

```
Resolving dependencies...
Got dependencies!
```

Note pub *does* resolve there, and does not fetch advisories. So resolving is not by itself the trigger.

Also ruled out: the lockfile is byte-identical across the committed copy, every local reproduction, and what the fragment records (`84edb5e8…`).

And withdrawing two earlier explanations of mine — the hook theory (hooks do not run during `pub get`) and the tool-bootstrap theory (`flutter_tools.stamp` matches the staged git HEAD exactly).

## So: measure rather than guess

Four hypotheses have looked sound and been wrong, and each cost a round. This adds a `do_pub_get_offline[prefuncs]` that records what pub actually sees — the pubspec and lock hashes, whether the pubspec is newer than the lock, whether the hooks section is present, what is in the app's `.dart_tool`, and how many files are really in `PUB_CACHE` and the SDK cache.

Scoped to this recipe so master's class stays clean, and marked for removal once the cause is known.

jwinarske:
The diagnostic did its job. Everything was in place except one thing:

```
pubspec/lock: 18aa3711…  pubspec.yaml    84edb5e8…  pubspec.lock
pubspec newer than lock: yes
hooks in pubspec: 1
app .dart_tool:
cache files: 23068, dirs: 4205
archive staged: 154 files
sdk cache files: 23037
```

Read that against the theories it kills:

- the lock **is** the one the fragment records (`84edb5e8…`), so `do_install_pubspec_lock` worked
- the hooks section **is** present, so the injection ran
- the cache holds **23,068 files** including `archive` with 154 of its own, so `do_seed_pub_cache` worked and the SDK cache staged correctly
- the app's `.dart_tool` is empty, so there is no stale resolution state

The only thing wrong is the mtime ordering. **A pubspec newer than its lockfile is one of the conditions pub uses to decide the lock is stale** — so it resolves, and a resolve fetches security advisories, which needs the network.

`do_install_pubspec_lock` was ordered `after do_unpack do_patch`. But a recipe that edits `pubspec.yaml` orders itself `before do_archive_pub_cache` — which is what the appstream_dart app does with its hooks section — and this class makes that task `noexec` without removing it from the graph. Ordering against it is what puts the lock install after the edit. `do_pub_get_offline` already did; `do_install_pubspec_lock` was missed when I fixed the other one.

I tested the mtime condition locally in isolation earlier and it passed, which is why I dropped the thread. That test was against a warm SDK on a machine where pub had other reasons not to re-resolve; it did not prove what I took it to prove.

Fix pushed here to get a verdict in one round. It is a class change and belongs in its own PR once green — I will split it rather than land it through this one.

jwinarske:
The mtime fix worked and was not the cause.

```
pubspec newer than lock: no      <- was "yes", now corrected
lock matches the fragment        <- 84edb5e8…
hooks in pubspec: 1
app .dart_tool: (empty)
cache files: 23068, archive staged: 154 files, sdk cache files: 23037
```

Every measurable input is now correct, and pub still resolves:

```
uri=https://pub.dev/api/packages/archive/advisories
```

## I should stop guessing

That is five hypotheses in a row — the hook, the tool bootstrap, the SDK cache seeding, the layer parse, the mtime ordering — each of which looked sound, each of which cost a CI round, and each of which was wrong or only partly right. Three of them turned out to be artifacts of my own environment rather than facts about CI. The ordering fixes were real improvements and worth keeping, but none of them was *the* cause, and I have no sixth theory I would back.

What I cannot do is reproduce it. With CI's staged flutter binary, CI's merged cache, every `.dart_tool` cleared, a fresh HOME and the network unshared away entirely, the same command succeeds locally every time. Something about the built environment differs in a way I have not been able to name from the outside.

## Recommendation

**Drop this app as the first opt-in.** I chose it for having three registered plugins and did not notice it also carries a path dependency on a package with a build hook and no lockfile of its own — the least representative app in the set, and the one whose resolution the lockfile is least likely to describe completely.

The mechanism itself is not in doubt:

- the fragment is complete and fetchable — `--runall=fetch` pulled all 156 tarballs in clean CI with checksums verified
- it resolves offline against a cold cache, proven repeatedly with the network unshared away
- `do_check_pubspec_lock`, `do_install_pubspec_lock` and `do_seed_pub_cache` all do what they claim, confirmed by the diagnostic

What is unproven is this app, and specifically whether a lockfile can describe an offline build for an app whose path dependency has its own unresolved dependency graph. That deserves its own issue rather than more rounds here.

## Concretely

1. Close this PR, or retarget it at an app without a path dependency
2. Split out the two ordering fixes — they are correct independently and the diagnostic proves the second one changed the state it was meant to
3. File the path-dependency-with-hook case as a known limitation of `pubvendor`

jwinarske:
Closing per the plan in the previous comment.

The ordering fix is split out as #859, where it stands on its own and its effect is measured rather than argued.

What this PR established, and which is worth not losing:

- the fragment is complete and fetchable — `--runall=fetch` pulled all 156 tarballs in clean CI with checksums verified
- it resolves offline against a cold cache, with the network unshared away entirely
- `do_install_pubspec_lock`, `do_check_pubspec_lock` and `do_seed_pub_cache` each do what they claim, confirmed by a diagnostic reading the state pub actually sees
- six defects fixed along the way, in #851, #853, #857, #858 and #859

What it did not establish is that **this** app builds offline. It was a poor choice for the first opt-in: it carries a path dependency on a package with a build hook and no lockfile of its own, which is the case a lockfile is least likely to describe completely. That limitation is worth its own issue rather than more rounds here.

Reopening is cheap if a better candidate app makes this branch useful again; otherwise the CI step shape (`--runall=fetch`, then `BB_NO_NETWORK=1`, then assert `_PluginRegistrant` in the packaged libapp.so) is the part to carry forward.

PR #856, opened 2026-08-30
ci: stop cloning meta-drm and meta-vulkan on master

jwinarske:
Neither contributes anything master CI builds, and parsing them is not free. meta-vulkan cost a full four-machine round on #855 when one of its recipes could not be parsed without the network — a failure with nothing to do with meta-flutter, in a layer nothing here depends on.

## What the references to them actually are

| reference | what it is |
|---|---|
| `filament-vk` in `flutter-auto_2.0`, `ivi-homescreen_2.0` | an **optional** `PACKAGECONFIG` dep, and only on the **2.0** recipes. CI builds v3, whose own comment records that *"filament-view was removed from the plugins tree"* |
| `swiftshader` in `flutter-engine_git.bb` | the copy **vendored inside the engine source** at `flutter/third_party/swiftshader`, not meta-vulkan's recipe. Zero `DEPENDS` on that recipe anywhere |
| `app-container-image-flutter-auto-vulkan` | genuinely needs meta-vulkan — and CI **never builds it** |
| meta-drm (`libdrmpp`, `libliftoff`, `kms-quads`, `drm-info`, `drm-howto`) | no meta-flutter recipe names any of them. ivi-homescreen v3 depends on `libdrm`, which is **oe-core**. `libliftoff` appears once, in a CHANGELOG |

## Why it is worth doing rather than just fixing the recipe

jwinarske/meta-vulkan#29 pinned those `SRCREV`s to shas and was worth landing on its own merits — a mutable tag as `SRCREV` can silently build different source. But it only fixed today's instance. Any recipe in either layer that cannot be parsed offline breaks the parse for the **whole build**, and we were paying that risk on every run for layers we do not use.

Same shape as the `meta-riscv` entry removed from this workflow earlier: a layer in the list without being needed.

## Kept deliberately

Both stay in the `rm -rf` so a persistent runner holding them from an earlier run does not keep a stale checkout — the same reason `poky` is still listed there. The comment now says so.

The other four branch workflows clone them too and can follow, once this is proven here.

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -85,7 +85,9 @@ jobs:
           # existing directory fails. Those failures are swallowed by
           # the backgrounded clones and bare 'wait', so an omitted
           # layer silently freezes at whatever revision it was first
-          # cloned at. poky is kept to clear any legacy checkout.
+          # cloned at. poky, meta-drm and meta-vulkan are kept here, though no
+          # longer cloned, so a runner that has them from an earlier run does
+          # not keep a stale checkout around.
           rm -rf poky bitbake openembedded-core meta-yocto \
                  meta-openembedded meta-drm meta-vulkan
           # 'wait' with no operands always returns 0, so a failed clone used to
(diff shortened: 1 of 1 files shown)


jwinarske:
This is independently good and can merge on its own; #855 is stacked on it, not the other way round.

**Its evidence is recorded on #855's run** (33325237558), because I cancelled #856's own round as redundant once they were stacked — that cancellation is why this PR's checks show red. In that run the parse error this fixes was **gone**: the build got past layer parsing and all the way to `do_pub_get_offline`, which it had never reached before. Dropping the two layers did not break anything; the build proceeded normally through the base recipes and the other apps on all three app-building machines.

Cancelling that round was a poor call on my part — it saved runner time and left this PR looking failed. If you would rather it carry its own green, re-running its checks after merging costs one round.

#855 should **not** merge — it hit a real limitation in the vendored cache for this particular app, unrelated to the layers.

Issue #861, opened 2026-08-31
roll: distinguish creating a missing lockfile from replacing a committed one


jwinarske:
`generate_pubcache_inc` deletes `pubspec.lock` and re-resolves unconditionally. That conflates two cases, and the fragment gives no way to tell them apart.

**No lockfile shipped.** Nothing to replace — the roll does once what the build already does every time (`common.inc:283`). Strictly better than today, where each builder resolves independently. More common than it looks: `flatpak-minimal-appstream-dart-flathub-catalog` commits no `pubspec.lock` at all.

**Lockfile shipped.** Deleting it substitutes our resolution for the author's, including any pin that avoided a bad version. Sometimes necessary given the pinned SDK — but it should be a visible decision.

## Proposed

- [ ] Create when absent; when present, try it first and re-resolve only if it fails against the pinned SDK
- [ ] Record which happened, in the roll output and the fragment header
- [ ] Per-entry override in `flutter-apps.json` for apps that must always re-resolve

## Documentation

`tools/pubvendor/README.md` says the roll "re-resolves the app's lockfile", quietly assuming there is one. And nothing in a generated `.inc` says whether it holds the author's resolution or ours. Whatever the behaviour, it needs to be in the README, the fragment header, and the roll's output.

Found while investigating #855; two claims I made there and in #852 about a lockfile mismatch were wrong for this reason — the app commits none, and what I measured was one pub had generated in an earlier build.

Related: #843, #852.

PR #862, opened 2026-08-31
appstream-dart-flathub-catalog: vendor the pub cache, and prove it in CI

### LINKED ISSUE #843: roll: generate the pub cache fragment for app recipes

jwinarske:
Closes #843. Retry of #855, now that the five class defects it uncovered have landed (#851, #857, #858, #859, #860). Same app, unchanged: the reason I gave for switching apps rested on a theory I later withdrew.

Dependencies arrive through `SRC_URI` with checksums instead of a networked `pub get`.

## The proof

```
bitbake --runall=fetch "$APP"      # network: populate DL_DIR from SRC_URI
BB_NO_NETWORK=1 bitbake "$APP"     # anything undeclared fails here
```

Offline is only half of it — `dart pub get` also resolves offline but writes no `.flutter-plugins-dependencies`, so an app can build green having registered nothing (#851, which no unit test could see). So the step also asserts the registrant reached the image:

```
strings -a "$LIBAPP" | grep -q '_PluginRegistrant'
```

This app registers three Linux plugins — `jni`, `path_provider_linux`, `url_launcher_linux` — so the assertion has something to catch.

The `dump_pub_state` prefunc from #855 is gone; #860 identified what it was there to find.

Note for review: this app commits no `pubspec.lock` upstream, so the vendored one is the roll's own resolution rather than the author's. That distinction is #861.


jwinarske:
First round failed, and it found something the unit tests could not: **the vendored packages were never being read.**

```
Because flathub_catalog depends on sqlite3 any which doesn't exist
(could not find package sqlite3 in cache), version solving failed.
```

`PUB_CACHE` was anchored to `WORKDIR`, but the fragment stages with `SRC_URI subdir=`, which unpacks into `UNPACKDIR` — `${WORKDIR}/sources` on this oe-core. Both directories exist in a workdir:

```
pub_cache/hosted/pub.dev          197 entries   <- PUB_CACHE, seeded from the SDK
sources/pub_cache/hosted/pub.dev   78 entries   <- the vendored packages, the fragment's exact count
```

`do_seed_pub_cache` fills the WORKDIR one from the SDK's `.pub-cache`, so it is not empty — which is why nothing complained. Resolution ran against the SDK's 174 packages, which carry `http` and `path` but not `sqlite3` or `drift`, and died on the first direct dependency the SDK does not ship.

The class already knew about this split; `do_install_pubspec_lock` takes `UNPACKDIR or WORKDIR`. The cache root just never got the same treatment.

Worth stating plainly: **every green result on this path so far proved nothing.** The SDK cache was standing in. An app whose dependencies the SDK happens to carry would have resolved green while vendoring nothing — which is exactly the class of "passes while inert" failure the `_PluginRegistrant` assertion exists to catch, and this one slipped past it.

So the fix is two parts. `PUB_CACHE_BASE` takes the same base as the lockfile, and `do_write_hosted_hashes` now asserts every declared package is present under `PUB_CACHE` before anything resolves. That loop already derives the exact path, so the check is free.

PR #865, opened 2026-08-31
master: make the staged SDK relocatable

jwinarske:
Every `package_config.json` in the SDK names its dependencies as absolute `file://` URIs under `${S}` — inside TMPDIR:

```
.dart_tool/package_config.json                            168
packages/flutter_tools/.dart_tool/package_config.json     101
dev/integration_tests/hook_user_defines/.dart_tool/...     51
```

A recipe consuming the SDK gets it under `recipe-sysroot-native`, where none of those paths exist. So the flutter tool decides `flutter_tools` is unresolved and re-runs `pub get` on it — **without `--offline`**:

```
executing: [.../example/flathub_catalog/] .../dart pub
  --directory .../recipe-sysroot-native/usr/share/flutter/sdk/packages/flutter_tools
  get --example
[+599797 ms] Running 1 shutdown hook
```

This has been happening on **every app build**. It went unnoticed because `do_compile` and `do_archive_pub_cache` have network, so the re-resolve just quietly succeeded and cost time. #862 is the first thing to run pub in a task without network, and there it does not fail — it hangs until killed.

`rootUri` may be relative to the file holding it; `flutter_tools`' own entry (`../`) already is. Making the rest relative makes the SDK relocatable, and removes an absolute TMPDIR from a staged artifact (#566).

## Verification

Ran the transform against the staged sysroot component:

| file | rewritten | outside SDK |
|---|---|---|
| `.dart_tool/package_config.json` | 168 | 0 |
| `packages/flutter_tools/.dart_tool/package_config.json` | 101 | 0 |
| `dev/integration_tests/hook_user_defines/.dart_tool/…` | 51 | 0 |

All **402** rootUri values resolve to a real directory afterwards, none dangling. `package_graph.json` carries no build paths and is untouched.

#862 stacks behind this — it cannot go green until this lands.

### DIFF
--- a/recipes-graphics/flutter-sdk/flutter-sdk_git.bb
+++ b/recipes-graphics/flutter-sdk/flutter-sdk_git.bb
@@ -168,6 +168,62 @@ do_install:append:class-target () {
     rm -rf ${D}${datadir}/flutter/sdk/bin/cache/dart-sdk/bin
 }
 
+# pub records where it resolved a package, and it records it absolutely: every
+# package_config.json in the SDK names its dependencies as file:// URIs under
+# ${S}, which is inside TMPDIR. Nothing recreates that path in a recipe that
+# consumes the SDK -- there it lives under recipe-sysroot-native -- so all of
+# those entries dangle, and the flutter tool concludes flutter_tools is
+# unresolved and re-runs `dart pub get` on it. That resolve is not offline. It
+# reaches pub.dev on every app build, and in a task with no network it does not
+# fail, it hangs until something kills it.
+#
+# The URIs may be relative to the file that holds them, which flutter_tools'
+# own entry ("../") already is. Making the rest relative makes the staged SDK
+# relocatable, and takes an absolute TMPDIR out of a staged artifact while it
+# is there. See #566.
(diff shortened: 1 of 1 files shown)


jwinarske:
This behavior was already observed and worked around — `master-build.yml:265`, on the integration-test step:

> each drives the flutter tool against its own `recipe-sysroot-native`, whose SDK files are hard-linked from the shared native sysroot, so concurrent writes under `bin/cache` are visible to the others. A recipe reading that directory mid-write decides the tool snapshot is stale and re-resolves `packages/flutter_tools` — which does not inherit our `--offline`, so in a network-isolated `do_compile` it can only fail.

The workaround there is to call bitbake once per recipe so the ten never build concurrently.

Same symptom, but the race is not the only way in. A single recipe building alone hits it too, because the `rootUri` entries dangle in *any* consuming sysroot regardless of what else is running. So the serialization in CI is treating one trigger of a condition that is always true.

If this holds, the integration tests could go back to a single `bitbake app1 app2 ...` call and build in parallel — worth trying separately once this lands, since it is ten sequential builds per round today.

Issue #867, opened 2026-08-31
roll: generate recipes for the Flutter SDK's own apps


jwinarske:
The SDK we already stage ships 47 buildable Flutter apps, and the layer builds none of them. They are pinned by `FLUTTER_SDK_TAG` and change with every roll, which makes them the cheapest broad regression surface available: no third-party repo to track, no SRCREV to bump, and they move in lockstep with the SDK under test.

Generate recipes for them as part of the roll, into `recipes-graphics/flutter-sdk`.

## What is there

Apps with `lib/main.dart` and a flutter SDK dependency, in the staged 3.47.1 SDK:

| tree | apps | with a `linux/` runner |
|---|---|---|
| `dev/` | 36 | 8 |
| `examples/` | 10 | 6 |
| `packages/` | 1 | 0 |

The obvious first set is `examples/` plus the handful under `dev/` that are not platform-specific:

```
examples/hello_world          examples/platform_channel
examples/layers               examples/platform_view
examples/multiple_windows     examples/texture
dev/integration_tests/flutter_gallery
dev/integration_tests/ui
dev/a11y_assessments          dev/manual_tests
```

Most of `dev/integration_tests` is Android- or iOS-specific and should stay out. A `linux/` runner is a reasonable proxy for desktop intent, though not a requirement — the layer builds `flutter build bundle`, not `flutter build linux`.

## Why this shape

They pair naturally with the ivi-homescreen integration tests in `recipes-graphics/toyota`, which share one `.inc` and one source revision. These would share `FLUTTER_SDK_TAG` the same way: one generated `.inc` carrying the common bits, one recipe per app, all rolling together.

Unlike `flutter-apps.json` entries there is no repo to clone — the source is already unpacked in the SDK the recipe depends on, so `S` points into `recipe-sysroot-native` and `SRC_URI` stays empty.

## Notes

- **None of the 47 ships a `pubspec.lock`** (0 of 47). So every one lands in the "create when absent" case from #861, and the roll would be generating the resolution rather than reproducing one. Worth settling #861 first, since this multiplies it by 47.
- Licensing is uniform — the SDK's own BSD-3-Clause — so `LIC_FILES_CHKSUM` can point at the staged `LICENSE` rather than being detected per app.
- Depends on #863: a roll that cannot fail should not be generating 47 recipes unattended.
[...]

PR #876, opened 2026-09-01
flutter-sdk: build the SDK's own example apps, and fix the rebuild that broke it

jwinarske:
Towards #867. The first SDK app, and the shape the other 46 can follow.

## No SRCREV, no manifest entry, no extra download

The source is the release archive `flutter-sdk-native` already fetches, named through `get_flutter_archive()` the same way. So the recipe follows `FLUTTER_SDK_TAG` automatically and shares the download — verified the URL is byte-identical to `flutter-sdk-native`'s and the tarball is already in `DL_DIR`.

That is simpler than #867 assumed. I had suggested there that whatever replaced the dead `flutter/flutter` manifest entry should take its revision from the pinned SDK; using the SDK's own archive means there is no revision to take.

## Why the whole tree unpacks

SDK apps declare `resolution: workspace` — they are members of a pub workspace rooted at the SDK, so an app copied out of it cannot resolve. `common.inc` already follows `.dart_tool/pub/workspace_ref.json` to the workspace root when building the snapshot, which is what makes this work at all.

The cost is real and worth naming: each such recipe unpacks the full SDK into its own `WORKDIR`. With `rm_work` only one or two exist at a time, but it is not free, and it is the main thing to weigh before scaling to 47.

## Layout

`recipes-graphics/flutter-sdk/apps/` rather than beside `flutter-sdk_git.bb`, which is one hand-maintained recipe that should not get buried as this family grows. `BBFILES` already globs three levels, so no `layer.conf` change.

## Verified before pushing

Parsed against the local build tree:

```
PN      flutter-sdk-examples-hello-world
PV      3.47.1
S       .../sources/flutter
SRC_URI .../stable/linux/flutter_linux_3.47.1-stable.tar.xz
FLUTTER_APPLICATION_INSTALL_SUFFIX  flutter-sdk-examples-hello-world
```

CI builds it on the three machines that can build Flutter apps.

## Not done here

Generating these from the roll. Hand-writing the first keeps "does this recipe shape work" separate from "how do we emit 47 of them" — the same order that worked for pubvendor in #862, where the shape turned out to need four fixes before anything built.

### DIFF
--- a/.github/workflows/master-build.yml
+++ b/.github/workflows/master-build.yml
@@ -357,6 +357,21 @@ jobs:
           # rounds reporting a failure the build did not have.
 
       #
+      # The first app that ships inside the Flutter SDK rather than in a
+      # repository of its own (#867). It has no source of its own at all: the
+      # SDK is already staged in recipe-sysroot-native, so the recipe copies
+      # the workspace out of it.
+      - name: Build flutter-sdk-examples-hello-world
+        # flutter build bundle has no linux-arm target platform,
+        # so the app recipes do not parse for a 32-bit machine
+        if: ${{ !cancelled() && steps.base.outcome == 'success' && matrix.machine != 'qemuarm' }}
+        shell: bash
+        working-directory: ../master-linux-dummy
+        run: |
+          . ./openembedded-core/oe-init-build-env
+          bitbake flutter-sdk-examples-hello-world
+
+      #
       # bluez_native: compiles a vendored sdbus-c++ from its own hook.
       - name: Build jwinarske-bluez-native-flutter-ble-scanner
         # flutter build bundle has no linux-arm target platform,
--- a/conf/include/common.inc
+++ b/conf/include/common.inc
@@ -487,6 +487,18 @@ def build_app(d, source_dir, flutter_sdk, pubspec_appname, flutter_application_p
         build_folder = os.path.join(source_root, 'build')
         if os.path.exists(build_folder):
             run_command(d, 'flutter clean -v', source_root, env)
+            # `flutter clean` removes .dart_tool, and the build below carries
+            # --no-pub whenever do_compile has no network, so nothing puts the
+            # resolution back. The snapshot is then handed a package_config.json
+            # that was deleted moments after do_compile resolved it. Resolve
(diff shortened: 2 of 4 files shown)


jwinarske:
Rewritten after three rounds. **Built end to end locally before this push** — all 2164 tasks, through packaging, QA and SPDX:

```
libapp.so        1,885,064 bytes
flutter_assets   9 entries
registrant       absent (correct — hello_world declares no plugins)
```

## This now carries a layer-wide fix, in its own commit

`common: re-resolve after flutter clean` is not about SDK apps. `do_compile` resolves, then `build_app` runs `flutter clean` whenever `build/` exists — deleting `.dart_tool` — and the build that follows carries `--no-pub` whenever `do_compile` has no network, so nothing puts the resolution back. The kernel snapshot is handed a `package_config.json` that was deleted moments after it was written.

**Only visible on a rebuild**, since a first build has no `build/` for the clean to trigger on. That is why every app has got away with it, and why my first CI round and my local reproduction disagreed with each other.

For a pub workspace member it is fatal rather than wasteful: `.dart_tool/pub/workspace_ref.json` goes with the directory, `common.inc:590` stops recognising the app as a member, and `--packages` falls back to a member `package_config.json` that a workspace member never has.

Worth reviewing on its own terms — it changes behaviour for every Flutter app in the layer, not just this recipe.

## What was wrong in the recipe (all mine)

| | |
|---|---|
| `S = "${WORKDIR}/..."` | `insane.bbclass:1450` makes anything outside UNPACKDIR fatal. My `bitbake -e` check only parsed; this fires in `do_unpack`. |
| no ordering vs `do_unpack` | `do_unpack[cleandirs]` is UNPACKDIR, so staging could be undone. Same unordered-sibling hazard as #857/#858/#859. |
| `pubspec.lock` not copied | A workspace member resolves against the **root** lockfile. Without it `--enforce-lockfile` fails outright. |
| `PUBSPEC_IGNORE_LOCKFILE = "1"` | Wrong here: the SDK ships a real, authored lockfile at the workspace root. Now `"0"`. |

The last two are the same mistake twice — I carried an assumption from the third-party apps into a case I had already proven was different.

